Issue #12, opened 2017-10-10
Clarify what an interface looks like


johnament:
The base of the spec should include rough details about what a rest client looks like.

derekm:
I'm using Jersey's `WebResourceFactory` with this patch: https://github.com/jersey/jersey/pull/235

I've achieved very successful results with this setup. The REST client I've built out with it exactly models the URL structures of the services I'm hitting (not a requirement of the setup I'm using, just a technique I'm using in organizing the resource interfaces).

I think it would be smart for MicroProfile REST Client leads to analyze this approach.

johnament:
@derekm that's useful to know.  I was under the impression that the jersey framework didn't support  proxying.  This is the approach we are considering.

johnament:
I'm adding the help wanted label to this.  I would like to see more people include sample interfaces of what they expect should be supported.  Note that each example should include a TCK test.

Might be good for early on contributors like @andymc12 and @derekm if interested :-)

johnament:
Based on some input, I think we should include some of the following:

- With and without `@Path` at the interface level
- Subresources
- BeanParams

johnament:
For the first bullet point, here's an example of the interface:

```java
public interface MyClient {
    @GET
    Response get();
}

MyClient client = RestCientBuilder.newBuilder().baseUrl("https://www.google.com").build(MyClient.class);
```

johnament:
For bullet point two:

```java
public interface MyClient() {
   @Path("/")
   Google google();
}

public interface Google {
   @GET
   Response get();
}
```

johnament:
bullet 3:

```java
public class MyBean {
   @QueryParam(name="query")
   private String language;
   @CookieParam(name="cookie")
   private String myCookie;
   @HeaderParam(name="Accepts")
   private String acceptsHeader;
// getters setters constructors
}

public interface MyClient{
    @GET
    Response get(@BeanParam MyBean myBean);
}
```

derekm:
Another feature the Jersey community is asking for is parameter converters, as seen in this PR to the Jersey codebase: https://github.com/jersey/jersey/pull/269

In Jersey-land, the BeanParam Support & Parameter Converters Support PRs are in conflict. If MP Rest Client could get ahead of these concerns & support both in one go, that would be huge for pushing the Java development experience forward.

PR #13, opened 2017-10-10
#12 Introducing basic spec information and client definition.

johnament:
Signed-off-by: John D. Ament <john.d.ament@gmail.com>

### DIFF
--- a/spec/src/main/asciidoc/clientexamples.asciidoc
+++ b/spec/src/main/asciidoc/clientexamples.asciidoc
@@ -1,0 +1,79 @@
+//
+// Copyright (c) 2017 Contributors to the Eclipse Foundation
+//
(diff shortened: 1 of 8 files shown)


spec/src/main/asciidoc/clientexamples.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+MicroProfile TypeSafe Rest Clients are defined as Java interfaces.
+
+=== Sample Definitions
+
+[source, java]
+----
+public interface MyServiceClient {
+    @GET
+    @Path("/greet")
+    Response greet();
```
pilhuhn: I think we should say that this is a fallback, but that the "real" functionality is rather the one shown below

PR #17, opened 2017-10-19
Adding a bootstrap API for programmatic creation of clients.

### LINKED ISSUE #2: Specify bootstrap API
Create a bootstrap API, not CDI specific, that allows a user to create a rest client proxy from an interface that has JAX-RS annotations on it.

johnament:
- Adding TCK tests
- Adding local unit tests

This closes #2

Signed-off-by: John Ament <john.d.ament@gmail.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.rest.client;
+
+import javax.annotation.Priority;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ServiceLoader;
+
+public abstract class RestClientBuilder {
+    public static RestClientBuilder newBuilder() {
+        ServiceLoader<RestClientBuilder> loader = ServiceLoader.load(RestClientBuilder.class);
+        List<RestClientBuilder> clientBuilders = new ArrayList<>();
+        loader.forEach(clientBuilders::add);
+        loader = ServiceLoader.load(RestClientBuilder.class, RestClientBuilder.class.getClassLoader());
+        loader.forEach(clientBuilders::add);
+
+        if(clientBuilders.size() == 0) {
+            throw new RuntimeException("No implementation of '"+RestClientBuilder.class.getSimpleName()+"' found");
+        }
+        clientBuilders.sort(Comparator.comparingInt(value -> {
+            Priority priority = value.getClass().getAnnotation(Priority.class);
+            if (priority == null) {
+                return 1;
+            }
+            else {
+                return priority.value();
+            }
+        }).reversed());
+        return clientBuilders.get(0);
+    }
+
+    public abstract RestClientBuilder baseUrl(URL url);
+
+    public abstract <T> T build(Class<T> clazz);
+}
--- a/spec/src/main/asciidoc/programmatic_lookup.asciidoc
+++ b/spec/src/main/asciidoc/programmatic_lookup.asciidoc
@@ -1,0 +1,36 @@
+//
+// Copyright (c) 2017 Contributors to the Eclipse Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+[[programmatic_lookup]]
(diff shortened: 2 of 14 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+package org.eclipse.microprofile.rest.client;
+
+import javax.annotation.Priority;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ServiceLoader;
+
+public abstract class TypeSafeClientBuilder {
```
pilhuhn: I think I don't really like the name of the Builder. One one side it is good that it is not the usual ClientBuilder that everyone has.
I wonder if it would make sense to just have a MPClientBuilder, that can then provide type safe and unsafe clients

spec/src/main/asciidoc/programmatic_lookup.asciidoc
```diff
@@ -0,0 +1,36 @@
...
+
+Type Safe Rest Clients support both programmatic look up and CDI injection approaches for usage.  An implementation of MicroProfile Rest Client is expected to support both use cases.
+
+=== Sample Builder Usage
+
+[source, java]
+----
+public class SomeService {
+   public Response doWorkAgainstApi(URL apiUrl, ApiModel apiModel) {
+       RemoteApi remoteApi = TypeSafeClientBuilder.newBuilder()
```
pilhuhn: I think from looking at this example that the TSCB should actually be a TypeSafeClient.
Or proxy() should be renamed build() or similar.

api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
pilhuhn: Yes, `RestClientBuilder` would work for me

api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+        clientBuilders.sort(Comparator.comparingInt(value -> {
+            Priority priority = value.getClass().getAnnotation(Priority.class);
+            if (priority == null) {
+                return 1;
+            }
+            else {
+                return priority.value();
+            }
+        }).reversed());
```
andymc12: Is the intent that higher value in the @Priority wins? If so, that is inconsistent with other uses of @Priority, particularly JAX-RS where higher priority actually has a lower value - i.e. @Priority(1) wins over @Priority(2).

spec/src/main/asciidoc/programmatic_lookup.asciidoc
andymc12: I agree that build() is better than proxy() - the user doesn’t really need to know or care that the underlying implementation is a dynamic proxy.

johnament: Does `build()` still take the interface, or how does the interface go in?  The idea I had behind it is that you could chain a reusable client, e.g.:

```java
TypeSafeClientBuilder builder = TypeSafeClientBuilder.newBuilder().baseUrl(..);
builder.build(Interface1.class);
builder.build(Interface2.class);
```

Does that still make sense?

api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.ServiceLoader;
+
+public abstract class TypeSafeClientBuilder {
+    public static TypeSafeClientBuilder newBuilder() {
+        ServiceLoader<TypeSafeClientBuilder> loader = ServiceLoader.load(TypeSafeClientBuilder.class);
+        List<TypeSafeClientBuilder> clientBuilders = new ArrayList<>();
+        loader.forEach(clientBuilders::add);
+        loader = ServiceLoader.load(TypeSafeClientBuilder.class, TypeSafeClientBuilder.class.getClassLoader());
```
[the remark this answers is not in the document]
johnament: The first call, `ServiceLoader.load(TypeSafeClientBuilder.class)` uses TCCL under the hood, this is the native impl (from the JDK):

```java
    public static <S> ServiceLoader<S> load(Class<S> service) {
        ClassLoader cl = Thread.currentThread().getContextClassLoader();
        return ServiceLoader.load(service, cl);
    }
```

api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
johnament: In the case of JAX-RS, multiple filters will apply.  The use of the highest priority wins is consistent with the CDI spec, when choosing observer method order, alternatives, etc.  So yes, the intent is that the highest priority impl will win (granted this is just to handle the case when multiple are on the classpath for some odd reason)

api/src/main/java/org/eclipse/microprofile/rest/client/TypeSafeClientBuilder.java
kenfinnigan: +1 to the name

spec/src/main/asciidoc/programmatic_lookup.asciidoc
kenfinnigan: +1 to build() instead of proxy()

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+import javax.annotation.Priority;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ServiceLoader;
+
+public abstract class RestClientBuilder {
+    public static RestClientBuilder newBuilder() {
```
Emily-Jiang: I'll prefer newInstance() or instance()

johnament: `instance()` tends to imply singleton, which isn't the case here.  I'd prefer to keep it more aligned to the actual instance name.  @pilhuhn @andymc12 @kenfinnigan thoughts?

andymc12: I like newBuilder() - it is consistent with the JAX-RS Client API.

Issue #15, opened 2017-10-12
Add OSGi/bnd support


johnament:
@Emily-Jiang  would you or someone else who's more familiar with it be able to help with OSGi support as needed by OpenLiberty?

Emily-Jiang:
@johnament in the api module's pom.xml, we need to build an OSGi bundle and properly version it. I'll create a PR for you. Each package needs to have a package-info.java.

PR #19, opened 2017-10-27
Define how providers are registered

### LINKED ISSUE #7: Message Body Reader/Writer support
Explain how readers/writers are wired in and used.

johnament:
Clarify how providers are registered, both via the builder API and at definition levels.  This currently covers the base providers available to a caller, which are known to work within the Client API.

This closes #7 

In addition, I'd like to get some input from @andymc12 , @reta and @sberyozkin on how true my claims about the client already supporting these providers is.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -17,13 +17,29 @@
 package org.eclipse.microprofile.rest.client;
 
 import javax.annotation.Priority;
+import javax.ws.rs.core.Configurable;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.ServiceLoader;
+import java.util.function.ToIntFunction;
 
-public abstract class RestClientBuilder {
+/**
+ * <p>
+ * This is the main entry point for creating a Type Safe Rest Client.  Implementations are expected
+ * to implement this class and register a service provider via {@link ServiceLoader} that works within their implementation.
+ * <p>
+ * Invoking <pre>RestClientBuilder.newBuilder()</pre> is intended to always create a new instance, not use a cached version.
+ * <p>
+ * If multiple implementations of RestClientBuilder are discovered, the one with the highest value of the {@link Priority} annotation is used
+ * <p>
+ * The {@link ServiceLoader} will first search via the current Thread's Context ClassLoader, then {@link RestClientBuilder}'s {@link ClassLoader}
+ * <p>
+ * The <pre>RestClientBuilder</pre> is a {@link Configurable} class as defined by JAX-RS.  This allows a user to register providers,
+ * implementation specific configuration.
+ */
+public abstract class RestClientBuilder implements Configurable<RestClientBuilder>{
     public static RestClientBuilder newBuilder() {
         ServiceLoader<RestClientBuilder> loader = ServiceLoader.load(RestClientBuilder.class);
         List<RestClientBuilder> clientBuilders = new ArrayList<>();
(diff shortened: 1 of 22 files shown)


johnament:
Note - I also did a bunch of javadoc clean up that I had neglected in the prior PR.

andymc12:
Overall, I think this looks really good.  My only comment (not a stopping point, just something to consider) is the `@RegisterProvider` annotation.  I think that users will prefer this approach over manually registering providers using the `RestClientBuilder`.  But the down-side is that the user would need to have the MP Rest Client APIs on the server side too - and the providers that the client wants to register may not even exist on the server (especially in the case of the client-side filters).  

Sorry if I missed this earlier, but would the MP Rest Client support inheritance like this?
```
@Path("/somePath")
public interface ServerSideResource {
...
}
```
```
@RegisterProvider(ClientSpecificProvider.class)
public interface ClientSideResource extends ServerSideResource { }
```
```
ClientSideResource client = RestClientBuilder.newBuilder().build(ClientSideResource.class);
```

If so, we can avoid tangling dependencies between the server and client, but it might also present other problems - like what does the proxy do if the sub-interface adds new methods that are not implemented on the server side resource... 

Sorry for the stream of consciousness - it must be Friday evening.  :)  Anyways, just something to think about - maybe we can talk more about this on Monday's call.

Thanks, Andy

johnament:
Hi Andy, great call outs.

1. Yes, I would expect inheritance to be supported in the use case you have.  When we start adding in more TCK tests we should include that test.  That approach as you mention would solve most of the problems you've noted.
2. I expect that the rest client is available server side, especially for gateway based applications that are just making other API calls.  E.g. while this can be used standalone, it's also expected to be available in a MP runtime.

One point, and maybe I've done this differently.  I don't typically share REST interfaces between client and server.  I may use different models, different representations client and server.  I've also used this approach to work between Java EE and Spring MVC applications.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+
+The `ReaderInterceptor` interface is a listener for when a read occurs against the response received from a remote service call.
+
+=== WriterInterceptor
+
+The `WriterInterceptor` interface is a listener for when a write occurs to the stream to be sent on the remote service invocation.
+
+== Provider Declaration
+
+In addition to defining providers via the client definition, interfaces may use the `@RegisterProviders` annotation to define classes to be registered as providers in addition to providers registered via the `RestClientBuilder`
```
pilhuhn: Can we clarify better what "in addition" is ?
Does the annotation augment what is passed in from Configurable or does it override?

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -17,13 +17,29 @@
...
+ * Invoking <pre>RestClientBuilder.newBuilder()</pre> is intended to always create a new instance, not use a cached version.
+ * <p>
+ * If multiple implementations of RestClientBuilder are discovered, the one with the highest value of the {@link Priority} annotation is used
+ * <p>
+ * The {@link ServiceLoader} will first search via the current Thread's Context ClassLoader, then {@link RestClientBuilder}'s {@link ClassLoader}
+ * <p>
+ * The <pre>RestClientBuilder</pre> is a {@link Configurable} class as defined by JAX-RS.  This allows a user to register providers,
+ * implementation specific configuration.
+ */
+public abstract class RestClientBuilder implements Configurable<RestClientBuilder>{
```
Emily-Jiang: This class needs to have setBuilder for it to be worked in OSGi environment, as noted in one of the builder pattern issue.

johnament: @Emily-Jiang understood, we are tracking your builder pattern concerns in a separate ticket.

kenfinnigan: I confess to not understanding requirements for OSGI.

However, it seems odd that we're having to add methods specifically for OSGI here, but `ClientBuilder` in JAX-RS which is also an abstract class doesn't have such methods.

@Emily-Jiang, how do Java EE 7/8 APIs work in OSGI without similar methods and bnd configuration?

johnament: I'm curious about this as well, but very willing to support OSGi.

When i look around the Open Liberty code, I found the following of interest:

https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.websphere.javaee.jaxrs.2.1/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.jaxrs.2.0.client/src/com/ibm/ws/jaxrs20/client/JAXRSClientBuilderImpl.java

However, JAX-RS itself is using `FactoryFinder` to locate the service... so I'm wondering if we just use that will it solve the problem? http://grepcode.com/file/repo1.maven.org/maven2/javax.ws.rs/javax.ws.rs-api/2.0/javax/ws/rs/client/FactoryFinder.java?av=f

Emily-Jiang: See what MP Config ConfigProviderResolver.setInstance(). The reason is that service loader pattern is not supported in OSGi. OSGi directly uses service discovery to discover the service of ConfigProviderResolver. The same thing is for MicroProfile Health Check. Which JavaEE 7/8 APIs did you see with a service loader pattern? Bnd configuration is to build a proper bundle to be used in OSGi. Some JavaEE 7/8 did not manage to build a proper bundle nor have their version correctly done. We should not follow the wrong path. Instead, MicroProfile releases should work well with OSGi for our future modularity story. 

By the way, in my recent talk of MP Config in OSGi, I was trying to promote MP programming model to be adopted by OSGi Alliance.

kenfinnigan: @Emily-Jiang I'm just trying to understand why OSGI is requiring certain things that are not part of Java EE APIs, as I have zero understanding of OSGI. For instance, the JAX-RS APIs in Java EE 7.

Are you saying that the JAX-RS API doesn't do enough from an OSGI perspective? Are there outstanding issues that haven't been resolved for this?

I'm not trying to block OSGI support. I'm just trying to understand what's actually required, as it seems to be adding things to the /api side that I wouldn't necessarily have expected.

With respect to the OSGI Maven dependency. Can that be marked as "optional" so that it doesn't become a transitive dependency of non OSGI containers?

johnament: While I understand the need, and agree that we need to continue discussing it, the usage of OSGi is not relevant to this PR.  Please continue the discussion within https://github.com/eclipse/microprofile-rest-client/issues/15

OndrejM: Is there any relevant reason why RestClientBuilder needs to be an abstract class? It would work equally fine as an interface which wouldn't force inheritance. See #28

Issue #4, opened 2017-07-07
Integration with config spec


johnament:
Once CDI has been declared, explain how to bootstrap with the config spec in place, matching on config keys and how to potentially override the client.

johnament:
Leverage config for CDI inject and potential bootstrap API support.
- Built in config keys for connection timeout/retry
- Built in config prefix, e.g. `rest.client...`

```java
@Inject
@RestClient("url.to.service") // does url.to.service become a config property?
private SomeInterface someInterface
```

Leave as stretch for 1.0.  Evaluate later if something isn't working quite right.

PR #22, opened 2017-10-28
Adding CDI support, as well as MicroProfile Config support.

### LINKED ISSUE #3: Integration with CDI
Explain how to bootstrap a client that is just an interface into a CDI bean, without requiring custom producers.

johnament:
This closes #3 & #4 

I went ahead and combined these two tickets.  There were two reasons for it:

- I didn't want to introduce a hard coded way to have a URL set, it should always be configured.
- Most of the configuration defined is optional, since they can be defined directly on the interface.

Any thoughts on this approach @andymc12 or @reta ?

### DIFF
--- a/spec/src/main/asciidoc/cdi.asciidoc
+++ b/spec/src/main/asciidoc/cdi.asciidoc
@@ -1,0 +1,49 @@
+//
+// Copyright (c) 2017 Contributors to the Eclipse Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+[[restcdi]]
+== MicroProfile Rest Client CDI Support
+
+Rest Client interfaces may be injected as CDI beans.  It is only required to support injection points that are defined by the application, programmatic lookup is not required to work.
+
+Interfaces are assumed to have a scope of `@Dependent` unless there is another scope defined on the interface.
+
+=== Support for MicroProfile Config
+
+For CDI defined interfaces, it is possible to use MicroProfile Config properties to define additional behaviors of the rest interface.  Assuming this interface:
+
+[source, java]
+----
+package com.mycompany.remoteServices;
+
+public interface MyServiceClient {
+    @GET
+    @Path("/greet")
+    Response greet();
+}
+----
+
+The values of the following properties will be provided via MicroProfile Config:
+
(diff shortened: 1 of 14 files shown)


spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -0,0 +1,49 @@
...
+    @Path("/greet")
+    Response greet();
+}
+----
+
+The following properties will then be available:
+
+- `com.mycompany.remoteServices.MyServiceClient.url`: The base URL to use for this service, the equivalent of the `baseUrl` method.  This property is considered required, however implementations may have other ways to define these URLs.
+- `com.mycompany.remoteServices.MyServiceClient.scope`: The fully qualified classname to a CDI scope to use for injection, defaults to `javax.enterprise.context.Dependent` as mentioned above.
+- `com.mycompany.remoteServices.MyServiceClient.providers`: A comma separated list of providers to include in the client, the equivalent of the `register` method or the `@RegisterProviders` annotation.
```
reta: I think it is implied, but may be better to be explicit about features as well, for example, `.. list of providers and features ...`

johnament: ack'd, will include in the next update.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -0,0 +1,49 @@
...
+package com.mycompany.remoteServices;
+
+public interface MyServiceClient {
+    @GET
+    @Path("/greet")
+    Response greet();
+}
+----
+
+The following properties will then be available:
```
Emily-Jiang: Do you mean the MyServiceClient provides the following properties? It will be a config source. If so, what is the ordinal for this config source, as these properties might be overridden by other config sources? The ordinal should be configurable as well.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -0,0 +1,49 @@
...
+
+- `com.mycompany.remoteServices.MyServiceClient.url`: The base URL to use for this service, the equivalent of the `baseUrl` method.  This property is considered required, however implementations may have other ways to define these URLs.
```
johnament: Based on call discussion, we're going to use a suffix based approach for the properties.  So they will instead look like `com.mycompany.remoteServices.MyServiceClient/mp-rest/url` and likewise for `providers` and `scope`

spec/src/main/asciidoc/cdi.asciidoc
johnament: Based on the discussion on the call, will rephrase this to be `The values of the following properties will be provided via MicroProfile Config`:

spec/src/main/asciidoc/cdi.asciidoc
johnament: We're going to create a follow up ticket for JAX-RS Features.

spec/src/main/asciidoc/cdi.asciidoc
Emily-Jiang: It fits better with the naming convention of Fault Tolerance config properties.

johnament:
I was thinking about this a bit more.  I think there's a bit more, wanted to put this out there.  Should we require a qualifier on the injection point, to indicate that this is meant to be the remote client?  If we don't, it becomes difficult to differentiate what is meant to be the rest client and other use cases.

johnament:
We'll add a new qualifier, name TBD, that represents that this interface should be a rest client.  The qualifier should not be required on the injection point, only if there are both client and server use cases.

Issue #21, opened 2017-10-27
Create separate API/SPI components


Emily-Jiang:
Create separate API and SPI.  The current `RestClientBuilder` represents an API that end users interact with.  Create a new SPI that implementers will satisfy that can be used in OSGi and non-OSGi aware environments.

Below is an agreed upon list of changes:
- Create a new `RestClientBuilderProvider` abstract class in a new `spi` package:
  - It should have the logic of doing the `ServiceLoader` based look up
  - Exposes static method that creates a builder
  - Have a static method that sets the provider
  - Have an instance level method to create a builder
  - The static method for creating a new builder should load the service if not already loaded, and call the instance method on that loaded service.  If set provider was called, it should use that instance instead of a SerivceLoader look up.
- Migrate `RestClientBuilder` to be an interface
  - It has a static method that just delegates down to the SPI to avoid end users calling the SPI

andymc12:
I prefer `newBuilder` because it is consistent with the JAX-RS Client API.  It also indicates that the returned builder will be a new instance.  I don't know how MP Config's builder works, but usually when I think of `instance()`, I think it returns the same instance each time it is called.  For MP Rest Client, I think we want a new instance for each call.

kenfinnigan:
+1 to `newBuilder`

johnament:
It was agreed in a prior call that we would keep `newBuilder` since it aligns better with the JAX-RS client API.  Likewise in MP Config, we use `Config.getConfig()` not `Config.instance()`

Emily-Jiang:
In the ConfigProviderResolver, we have instance() not newResolver() and then it has getBuilder()

Emily-Jiang:
Please refer to the config and health apis. The pattern is: Provider/Resolver, Builder and final object. There should be a method to set Provider/Resolver for OSGi environment. I suggest to create a Resolver class, you can discover the builder and then say newBuilder() and setBuilder() for OSGi environment.

johnament:
Discussed on a call today, we're going to create a new class `RestClientBuilderProvider` which will handle the state of the builder provider and manage OSGi requirements.  Convert `RestClientBuilder` into an interface with a static method for create a newBuilder to avoid end users needing to deal with the SPI.

PR #28, opened 2017-11-01
Make RestClientBuilder an interface instead of an abstract class

OndrejM:
RestClientBuilder would work equally well as an interface instead of abstract class. An abstract class forces inheritance and isn't necessary since we can use Java 8 default methods to add default behavior if needed. Private code is in package visible static helper class.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -26,24 +26,17 @@ import java.util.function.ToIntFunction;
 
 /**
  * <p>
- * This is the main entry point for creating a Type Safe Rest Client.
- * Implementations are expected to implement this class and register a service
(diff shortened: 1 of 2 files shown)


johnament:
HI @OndrejM  the plan was actually to do the conversion to an interface + introduce a provider class.  I have to update #21 to better describe this.  Its to make the OSGi use cases work cleaner.

OndrejM:
Great, @johnament. You can use my code or I can reraise it against master after your PR is merged.

johnament:
@OndrejM I've updated the ticket.  If you want to take a look and see if you have time finish it up.  Its basically moving the static method's contents into an abstract SPI class, adding a set provider method and only looking up the provider once.

Emily-Jiang:
In general, I think we are heading towards the same direction. sounds good.

OndrejM:
I'm merging this PR and will do another PR for #21

OndrejM:
I realized this was raised against a PR branch. I'll reraise it against master.

PR #32, opened 2017-11-05
Define how to map an HTTP response code into an exception.

### LINKED ISSUE #8: Inverse ExceptionMappers
JAX-RS defines an ExceptionMapper that converts an Exception into a Response object.

We should be able to define a mapping of the inverse, by the HTTP response code, what exception to throw.

johnament:
This closes #8

Signed-off-by: John Ament <john.d.ament@gmail.com>

Not sure if you have any input @andymc12 or @reta

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
@@ -29,6 +29,9 @@ public interface ResponseExceptionMapper {
 
     /**
      * Converts a given Response into a Throwable.  The runtime will throw this if it is present.
+     *
+     * If this method reads the response body as a stream it must ensure that it resets the stream.
+     *
      * @param response the JAX-RS response processed from the underlying client
      * @return A throwable, if this mapper could convert the response.
      */
@@ -37,6 +40,8 @@ public interface ResponseExceptionMapper {
     /**
      * Whether or not this mapper will be used for the given response.  By default, any response code of 400 or higher will be handled.
      * Individual mappers may override this method if they want to more narrowly focus on certain response codes.
+     * 
+     * If this method reads the response body as a stream it must ensure that it resets the stream.
      *
      * @param response
      * @return
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.rest.client.ext;
+
+import javax.annotation.Priority;
+import javax.ws.rs.core.Response;
+import java.util.Optional;
+
+/**
+ * Converts an JAX-RS Response object into an Exception.
+ *
+ */
+public interface ResponseExceptionMapper {
+    int DEFAULT_PRIORITY = 100;
+
+    /**
+     * Converts a given Response into a Throwable.  The runtime will throw this if it is present.
+     * @param response the JAX-RS response processed from the underlying client
+     * @return A throwable, if this mapper could convert the response.
+     */
+    Optional<Throwable> toThrowable(Response response);
+
+    /**
+     * Whether or not this mapper will be used for the given response.  By default, any response code of 400 or higher will be handled.
+     * Individual mappers may override this method if they want to more narrowly focus on certain response codes.
+     *
+     * @param response
+     * @return
+     */
+    default boolean handles(Response response) {
+        return response.getStatus() >= 400;
+    }
+
+    /**
+     * The priority of this mapper.  By default, it will use the {@link Priority} annotation's value as the priority.
+     * If no annotation is present, it uses a default priority of 100.
+     * @return
+     */
+    default int getPriority() {
+        Priority priority = getClass().getAnnotation(Priority.class);
+        if(priority == null) {
+            return DEFAULT_PRIORITY;
+        }
+        return priority.value();
+    }
+}
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
@@ -17,30 +17,30 @@
 package org.eclipse.microprofile.rest.client.ext;
 
 import javax.annotation.Priority;
+import javax.ws.rs.Priorities;
 import javax.ws.rs.core.Response;
-import java.util.Optional;
 
 /**
  * Converts an JAX-RS Response object into an Exception.
  *
  */
 public interface ResponseExceptionMapper {
-    int DEFAULT_PRIORITY = 100;
+    int DEFAULT_PRIORITY = Priorities.USER;
 
     /**
-     * Converts a given Response into a Throwable.  The runtime will throw this if it is present.
+     * Converts a given Response into a Throwable.  The runtime will throw this if it is non-null.
      *
      * If this method reads the response body as a stream it must ensure that it resets the stream.
      *
      * @param response the JAX-RS response processed from the underlying client
      * @return A throwable, if this mapper could convert the response.
      */
-    Optional<Throwable> toThrowable(Response response);
+    Throwable toThrowable(Response response);
 
     /**
      * Whether or not this mapper will be used for the given response.  By default, any response code of 400 or higher will be handled.
      * Individual mappers may override this method if they want to more narrowly focus on certain response codes.
-     * 
+     *
      * If this method reads the response body as a stream it must ensure that it resets the stream.
      *
      * @param response
@@ -52,7 +52,7 @@ public interface ResponseExceptionMapper {
 
     /**
      * The priority of this mapper.  By default, it will use the {@link Priority} annotation's value as the priority.
-     * If no annotation is present, it uses a default priority of 100.
+     * If no annotation is present, it uses a default priority of {@link Priorities.USER}.
      * @return
      */
     default int getPriority() {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Converts an JAX-RS Response object into an Exception.
+ *
+ */
+public interface ResponseExceptionMapper {
+    int DEFAULT_PRIORITY = 100;
+
+    /**
+     * Converts a given Response into a Throwable.  The runtime will throw this if it is present.
+     * @param response the JAX-RS response processed from the underlying client
```
OndrejM: Please add the following from the spec to the javadoc too:
> If this method reads the response body as a stream it must ensure that it resets the stream.

OndrejM:
Looks good!

api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
johnament: Technically it should be both of the methods.  But yes, will add.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.annotation.Priority;
+import javax.ws.rs.core.Response;
+import java.util.Optional;
+
+/**
+ * Converts an JAX-RS Response object into an Exception.
+ *
+ */
+public interface ResponseExceptionMapper {
+    int DEFAULT_PRIORITY = 100;
```
andymc12: Instead of hardcoding to 100, can we set it so the the default priority is `Priorities.USER` (which is 5000) or `Priorities.USER + 1` (or something higher than USER) for container-provided ResponseExceptionMappers.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -47,6 +47,20 @@ The `ReaderInterceptor` interface is a listener for when a read occurs against t
 
 The `WriterInterceptor` interface is a listener for when a write occurs to the stream to be sent on the remote service invocation.
 
+=== ResponseExceptionMapper
+
+The `ResponseExceptionMapper` is specific to MicroProfile Rest Client.  This mapper will take a `Response` object retrieved via an invocation of a client and convert it to a `Throwable`, if applicable.  The runtime should scan all of the registered mappers, sort them ascending based on `getPriority()`, find the ones that can handle the given `Response`, and invoke them.  The first one discovered where `toThrowable` returns a non-empty `Optional` object will be thrown by the runtime.
```
johnament: We're going to leave this behavior.  Its implementation specific if it calls all handles methods before invoking the toThrowable, but its only required to call toThrowable and handles until it finds a working pair.  We will always call these methods to determine if a response requires an exception to be thrown and not pre-check on response codes.  

In addition, we will require the runtime to register a default mapper that has priority `Integer.MAX_VALUE` that will map any >= 400 response into a `WebApplicationException` with the `Response` object applied.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Converts a given Response into a Throwable.  The runtime will throw this if it is present.
+     *
+     * If this method reads the response body as a stream it must ensure that it resets the stream.
+     *
+     * @param response the JAX-RS response processed from the underlying client
+     * @return A throwable, if this mapper could convert the response.
+     */
+    Optional<Throwable> toThrowable(Response response);
```
kenfinnigan: Agreed in Monday's call that we would simply return `Throwable`

api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
johnament: @OndrejM its been added if you can look.

johnament:
@andymc12 @kenfinnigan if you can review whenever you get a chance.

spec/src/main/asciidoc/providers.asciidoc
andymc12: I think the text is out of sync with the code - the code just returns a Throwable, so instead of:

> The first one discovered where `toThrowable` returns a non-empty `Optional` object will be thrown by the runtime.

I think it should be something like:
The runtime will throw the first non-null Throwable returned from a mapper.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -47,6 +47,28 @@ The `ReaderInterceptor` interface is a listener for when a read occurs against t
...
+
+==== How to Implement ResponseExceptionMapper
+
+The specification provides default methods for `getPriority()` and `handles(Response)` methods.  Priority is meant to be derived via a `@Priority` annotation added to the client.  The runtime will sort ascending, taking the one with the lowest numeric value first to check if it can handle the `Response` object.  The usage of ascending sorting is done to be consistent with JAX-RS behavior.
```
andymc12: `Priority is meant to be derived via a @Priority annotation added to the client.`

should be:

`Priority is meant to be derived via a @Priority annotation added to the client's ResponseExceptionMapper class.`

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -47,6 +47,28 @@ The `ReaderInterceptor` interface is a listener for when a read occurs against t
...
+
+Likewise, the `handles` method by default will handle any response code >= 400.  You may override this behavior if you so choose to handle other response codes (both a smaller ranger and a larger range are expected).
```
andymc12: Optional: might want to add another reminder to buffer the entity if reading it in this method.  I suspect this might be a trouble area for some users.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -47,6 +47,28 @@ The `ReaderInterceptor` interface is a listener for when a read occurs against t
...
+
+The `toThrowable(Response)` method actually does the conversion work.  This method should not raise any `Throwable`, instead just return a `Throwable` if it can.  It returns `Optional<Throwable>` in case it just wants to do a payload inspection.
```
andymc12: Another out-of-sync with the code comment.  I don't remember which way we decided on the call - whether to use Optional or just have the user return null if they don't want to convert the response to an exception.  Either way sounds good to me, but the text and the code should align.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -47,6 +47,28 @@ The `ReaderInterceptor` interface is a listener for when a read occurs against t
...
+
+Any methods that read the response body as a stream must ensure that they reset the stream.
+
+==== Built in ResponseExceptionMapper
+
+Each implementation will provide out of the box a `ResponseExceptionMapper` implementation that will map the response into a `WebApplicationException` whenever the response status code is >= 400.  It has a priority of `Integer.MAX_VALUE`.  It is meant to be used as a fall back whenever an error is encountered.  This mapper will be registered by default to all client interfaces.
+
```
andymc12: Just a thought.... would it make sense to disable the built-in mapper if the user has registered their own?

Here is the scenario - the user considers only specific HTTP responses to be an exception (i.e. 401, 403, 500) but others are ok (i.e. 404 might just indicate that a key is missing from a hash map or something similar - not necessarily an exception).  They create and register one or more of their own custom mappers, but then they would also need to explicitly disable the built-in mapper to avoid getting a WebApplicationException when a non-exception HTTP response code is returned.

What do you think?  Other options include:
1) Don't automatically register the built-in mapper - but make it available in the spec APIs so that users could easily add it.
2) Make some kind of annotation-based switch to disable the built-in mapper, something like `@ReplacesBuiltIn` so that a user doesn't need to add any config.

Anyway, just something to think about - this should not hold up delivery.

spec/src/main/asciidoc/providers.asciidoc
johnament: Good catch.

spec/src/main/asciidoc/providers.asciidoc
johnament: You mean instead of the property?  The reasoning behind it is to make it simple to setup from an API consumer's point of view.  I'm assuming that the consumer is defining the interface, not the producer.  I'd rather see it used by default without a user having to do anything else.

andymc12: My thought was more along the lines of "if you (user) do not register a ResponseExceptionMapper, then the framework will provide a default one (>=400 --> WebApplicationException)", but "if you do register any mapper at all, then the framework will not provide the default one".

I'm starting to doubt whether my idea is actually all that intuitive though... I think it would mean less code/config for some cases, but might be more confusing for some customers.

andymc12:
What should happen in the case where a user registers a mapper for a declared exception (i.e. not a subclass of RuntimeException or Error) but then the client interface's method signature does not list that exception type on the throws clause?

For example:
```
    public MyException extends Exception { ... }
```

```
    public MyMapper implements ResponseExceptionMapper {
        @Override
        public Throwable toThrowable(Response r) {
            throw new MyException();
        }
        ...
```

```
    @Path("/someApi")
    public interface MyClient {
        @GET
        public String get(); /* does not declare any exceptions */
    }
```

The MP Rest Client implementation should throw a MyException, but can't since it is not declared.  In this case, should the implementation throw a WebApplicationException with MyException as the cause?  Or should the implementation skip MyMapper and choose a different mapper that returns a legally throwable Exception?  Or maybe the implementation should log an error and not throw anything at all?

What do you think?

kenfinnigan:
In that situation, should the deployment result in an exception on startup?

Informing the developer that the requested exception can't be thrown from that method?

johnament:
@andymc12 good call out.  I'm almost wondering, should we reduce the scope of the mapper to only be `RuntimeException`?  I'd rather not have a deployment issue when it can be dealt with programmatically.

andymc12:
I can be persuaded otherwise, but in general, I prefer declared exceptions over RuntimeExceptions - I think it makes the coding a lot easier (the contract is much clearer, etc.).  What do you think if we were to make the Exception type a generic on the ResponseExceptionMapper interface (similar to JAX-RS ExceptionMappers), then the container/implementor would be able to determine if the REM applied to the given client method based on it's signature.

So for some methods, like `@GET public Account getAccount(@PathParam("acctNum") int num) throws UnknownAccountException` the container would convert a specific response (i.e. 404) to throw the declared exception.  But for other methods that do not declare that it throws that exception, like `@POST public void createAccount(AccountInfo info)` the container would skip the "404-to-UnknownAccountException" and instead would throw the default WebApplicationException - indicating that this is not a problem with application or usage, but rather that the remote server doesn't have the resource at the expected location (which is likely a server problem).

What do you think?

johnament:
Andy asked to get this merged, to allow him to move forward on checked exceptions.  I didn't get to the TCK changes yesterday, but I'll plan for that when I get back next week.

PR #37, opened 2017-11-23
Checked exceptions

andymc12:
This PR adds text explaining how a user might handle checked exceptions - i.e. what to do if a ResponseExceptionMapper returns an exception that must be declared on the interface method but isn't (i.e. an exception type that does not extend RuntimeException or Error).  In that case, that exception is ignored, and the MP Rest Client implementation must attempt to throw the returned exception from the next highest-priority ResponseExceptionMapper.

This PR also parameterizes the ResponseExceptionMapper interface.  This should make it easier for implementations to determine whether the client method declares this type of exception.

It also adds a dependency on javax.inject required to build the CDI portions of the API.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
@@ -24,18 +24,19 @@ import javax.ws.rs.core.Response;
  * Converts an JAX-RS Response object into an Exception.
  *
  */
-public interface ResponseExceptionMapper {
+public interface ResponseExceptionMapper<T extends Throwable> {
     int DEFAULT_PRIORITY = Priorities.USER;
 
     /**
-     * Converts a given Response into a Throwable.  The runtime will throw this if it is non-null.
+     * Converts a given Response into a Throwable.  The runtime will throw this if it is non-null
+     * AND if it is possible to throw given the client method's signature.
      *
      * If this method reads the response body as a stream it must ensure that it resets the stream.
      *
      * @param response the JAX-RS response processed from the underlying client
      * @return A throwable, if this mapper could convert the response.
      */
-    Throwable toThrowable(Response response);
+    T toThrowable(Response response);
 
     /**
      * Whether or not this mapper will be used for the given response.  By default, any response code of 400 or higher will be handled.
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -34,6 +34,11 @@
             <artifactId>cdi-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+            <version>1</version>
+        </dependency>
+        <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
--- a/spec/src/main/asciidoc/providers.asciidoc
+++ b/spec/src/main/asciidoc/providers.asciidoc
@@ -49,7 +49,7 @@ The `WriterInterceptor` interface is a listener for when a write occurs to the s
 
 === ResponseExceptionMapper
 
-The `ResponseExceptionMapper` is specific to MicroProfile Rest Client.  This mapper will take a `Response` object retrieved via an invocation of a client and convert it to a `Throwable`, if applicable.  The runtime should scan all of the registered mappers, sort them ascending based on `getPriority()`, find the ones that can handle the given `Response`, and invoke them.  The first one discovered where `toThrowable` returns a non-null `Throwable` will be thrown by the runtime.
+The `ResponseExceptionMapper` is specific to MicroProfile Rest Client.  This mapper will take a `Response` object retrieved via an invocation of a client and convert it to a `Throwable`, if applicable.  The runtime should scan all of the registered mappers, sort them ascending based on `getPriority()`, find the ones that can handle the given `Response`, and invoke them.  The first one discovered where `toThrowable` returns a non-null `Throwable` that can be thrown given the client method's signature will be thrown by the runtime.
 
 ==== How to Implement ResponseExceptionMapper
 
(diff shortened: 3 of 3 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/ext/ResponseExceptionMapper.java
```diff
@@ -24,18 +24,19 @@
  * Converts an JAX-RS Response object into an Exception.
  *
  */
-public interface ResponseExceptionMapper {
+public interface ResponseExceptionMapper<T extends Throwable> {
```
johnament: By allowing the class to throw any `Throwable`, we're putting the checking logic into `handles`.  I'm not sure I understand why it's parameterized, other than to make it align to the signature of `ExceptionMapper`, which does make more sense since you have to consume the `Throwable` and casting it isn't useful.

andymc12: Yes, aside from aligning with `ExceptionMapper`, it should also make it easier for implementors (containers) to determine whether they can throw the exception returned from `toThrowable` without needing to invoke it.  The `handles` method only sees the Response object, so it does not know the method signature of the method being invoked - an alternative approach would be to pass the method signature (or at least the set of checked exceptions) to the `handles` method ; then the user could determine whether they should throw a particular exception or not. 

My thought is that if the implementor knows early enough that they cannot throw the specified exception type, then they can skip calling the `toThrowable` method.  

Take for instance a user registers a bunch of different mappers, such as:
* `@Priority(4500) MapperOne<MyException>` - throws user-created exceptions
* `@Priority(4600) MapperTwo<NullPointerException>` - throws NPEs 
* `@Priority(4700) MapperThree<Throwable>` - throws any number of exceptions, RuntimeExceptions, Errors, and "checked" exceptions
* `@Priority(5000) DefaultMapper<WebApplicationException>` - default mapper that throws WAEs on >=400

If the client method is `public Response get() throws MyException` then the implementor needs to check the `handles` and (assuming handles returns true) the `toThrowable` of all four mappers until it gets a non-null throwable.

If the client method is `public Response get2()` (no checked exceptions), then the implementor could skip the `handles` and `toThrowable` method for MapperOne because it knows that it could not throw that exception in this method.  It would still need to call those methods on the other mappers - and (unfortunately) it would need to check the type of exception returned from MapperThree (if it is a subclass of RuntimeException or Error, then it throws it - otherwise it must ignore the returned exception and move on to the DefaultMapper).

By avoiding the `toThrowable` method we avoid the expense of creating a new exception when we know we can't use it.

That said, I know that these changes will break the TCK changes you just submitted for review.  If you want, we can delay these changes until after your TCK changes are in.  Then I can modify the TCK in this PR.

johnament: Andy, I think if that's your perspective we're out of sync.  We should review at the next meeting, but my intention is that handles is always called to confirm if the `Response` is meant to throw an exception.  It should have very little to do with the type of exception thrown.  Don't forget that even if your method signature is `Response get() throws MyCheckedException` if a higher priority (lower value) mapper converts that to a `RuntimeException` that will be the exception thrown, even if the checked exception matches the signature.  Remember, one of the things we discussed last time was that if a checked exception is returned and the method does not declare one we must ignore it.  But that shouldn't cause the `handles` method to not be invoked.

andymc12: I don't think we are too far out of sync.  

> my intention is that handles is always called to confirm if the Response is meant to throw an exception.

I think this is the only place where we differ.  For the most part, I would hope that users will implement the `handles` method with simple, high-performing checks (like `return Response.getStatus() == 404`) - and in that case, I don't think there is a problem mandating that implementors always call the `handles` method.  It is the more complex cases that concern me (i.e. "buffer the response entity ; parse the JSON or XML, etc. response body to see if a particular element exists ; if so, return true").  

We could avoid these calls in cases where we know ahead of time that we won't be able to throw the returned exception.  

> Don't forget that even if your method signature is Response get() throws MyCheckedException if a higher priority (lower value) mapper converts that to a RuntimeException that will be the exception thrown, even if the checked exception matches the signature.

Yes, so implementors would essentially need to create an ordered list of mappers based on the priority.  Then for each mapper in the list (1) check to see if the mapper throws an exception that could be thrown in the client method.  If so, (2) check to see if that mapper's `handles` method returns true, and if so, (3) call the mapper's `toThrowable` method.  If the returned Throwable can be thrown in the client method, then we have a match and we are done.  If not, then we move on to the next mapper in the list.

> Remember, one of the things we discussed last time was that if a checked exception is returned and the method does not declare one we must ignore it. But that shouldn't cause the handles method to not be invoked.

I would prefer to avoid calling the `handles` method if we know we can't use that mapper, but I can see the benefit of always calling the `handles` method on every mapper until we have a match.  If we do that, then we could swap 1 and 2 above - we call `handles` first, then check the method signature - that still allows implementations to avoid the `toThrowable` call in cases where we know the method cannot throw the exception.

Thanks for the review feedback, Andy

johnament: So let's break this down into two sides of the issue I see.  1. The use of generics. 2. When should we filter out methods that don't work.

1. The use generics makes the code harder to use from a user's standpoint.  In addition, I don't believe as an implementor of this its going to make things work any better.  This is perfectly valid as an implementation:

```java
public class MyExceptionMapper implements ResponseExceptionMapper {
	public MyException toThrowable(Response r) {
	   // ...
	}
}
```

The caller of this still sees the satisfied interface, and we don't have to deal with the generic signature.  Since proxy invocations allow throwing of any `Throwable` throwing the found non-null value.  If we did want to invterrogate the return value, we can still read the method signature to determine if its a checked exception, but ultimately we need to check if the returned value is the right type of checked exception.  The same method returning a `Throwable` would need to have these checks applied as well.  Alternatively, I'm fine if the underlying exception thrown turns into a `UndeclaredThrowableException` in the case where a mapper returns a checked exception but the method signature doesn't declare the checked exception.

2. This PR starts to look like the use `handles` really is of little benefit, and I'm OK with that.  I agree with the concerns about reading the response body, when I added that note I said to myself "this is going to cause an issue."  So what if we remove the method all together, and just say that its up to the implementation to check if the `Response` matches some criteria they have, and that by reading the body they must reset it.  Otherwise, what if we just change `handles` to only take the status code & headers returned?

PR #38, opened 2017-11-25
Adding TCK tests for exception mapping and default mapper.

### LINKED ISSUE #8: Inverse ExceptionMappers

johnament:
This closes #8

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/CallMultipleMappersTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/CallMultipleMappersTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2017 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.rest.client.tck;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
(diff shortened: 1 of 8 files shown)


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/DefaultExceptionMapperTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @BeforeTest
+    public void resetHandlers() {
+        TestResponseExceptionMapper.reset();
+        TestResponseExceptionMapperOverridePriority.reset();
+        wireMockServer
+            .stubFor(get(urlEqualTo("/")).willReturn(aResponse().withStatus(STATUS).withBody(BODY)));
+    }
+
+    @Test
+    public void testWithOneRegisteredProvider() throws Exception {
```
andymc12: I would suggest a different name for this method - maybe something along the lines of `testPropagationOfResponseDetails`.  It seems like registering the other mapper is a good thing to test, but the _real_ test here is that the default mapper propagates the status code and response body.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/CallMultipleMappersTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @BeforeTest
+    public void resetHandlers() {
+        TestResponseExceptionMapper.reset();
+        TestResponseExceptionMapperOverridePriority.reset();
+        wireMockServer
+            .stubFor(get(urlEqualTo("/")).willReturn(aResponse().withBody("body is ignored in this test")));
+    }
+
+    @Test
+    public void testWithOneRegisteredProvider() throws Exception {
```
andymc12: This method should probably be renamed to `testWithTwoRegisteredProviders` - or else we should only register one (and remove line 66).

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/DefaultExceptionMapperTest.java
johnament: I'm going to go with the name `testPropagationOfResponseDetailsFromDefaultMapper` since I want to see that the `Response` object is propagated because of the default mapper, or is that implied by the class name?

PR #41, opened 2017-11-29
Fixes #21 - a separate SPI for providing RestClientBuilder instances

### LINKED ISSUE #21: Create separate API/SPI components

OndrejM:
I created a separate class `RestClientBuilderResolver` to provide instances of `RestClientBuilder`. I think `RestClientBuilderResolver` is a better name than `RestClientBuilderPovider` agreed in #21 because in the config spec, `ConfigProvider` is a part of API and `ConfigProviderResolver` is a part of SPI. The new `RestClientBuilderProvider` does a similar job to `ConfigProviderResolver. There's nothing like ConfigProvider in the REST client API because there's no single client API, only builder.

I also left the default behavior of loading `RestClientBuilder` instances via service loader but moved it to the new class to a method which can be overridden by the implementaions. The code is very simple and might be enough for some implementations.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -22,13 +22,18 @@ import org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver;
 /**
  * This is the main entry point for creating a Type Safe Rest Client.
  * <p>
- * Invoking {@link newBuilder()} is intended to always create a new instance, not use a cached version.
+ * Invoking {@link newBuilder()} is intended to always create a new instance,
+ * not use a cached version.
+ * </p>
  * <p>
- * The <code>RestClientBuilder</code> is a {@link Configurable} class as defined by JAX-RS.  This allows a user to register providers,
- * implementation specific configuration.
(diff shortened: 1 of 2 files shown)


OndrejM:
I'm thinking about creating a new git repo as a template for other repos. It could include example maven config for checkstyle, OSGi bnd, sonatype deployment but it could also include source code for frequent patterns like SPI with resolvers etc.

johnament:
@andymc12 @kenfinnigan could you guys take a look? Its the same pattern as config, so I think we're fine.

PR #43, opened 2017-11-30
Fixes failing test.

OndrejM:
The test also gives an example how to register builder resolver using both service loader and static setter


johnament:
@OndrejM ok, so there's definitely still an issue with your change.  Or maybe I didn't think through what we were doing, not sure.

Prior to your change, an implementor only needs to implement `RestClientBuilder`.  Now they need to implement the resolver class.  I suspect the methods in `RestClientBuilderResolver` should now become abstract, and up to the implementation on how to provide those.

johnament:
Oh, looks like its only the `newBuilder` method, that should become abstract.  `getBuilderPriority` should go away.

Issue #42, opened 2017-11-29
Replace handles method


johnament:
Replace the current `handles` method so that instead of the `Response` just pass in the response status code and the headers of the response.

johnament:
Done by @andymc12

PR #50, opened 2017-12-05
Use status and headers instead of Response in handles() method (#42)

andymc12:


andymc12:
This resolves issue #42

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ExceptionMapperTest.java
```diff
@@ -52,7 +52,9 @@
     public void resetHandlers() {
         TestResponseExceptionMapper.reset();
         TestResponseExceptionMapperOverridePriority.reset();
-        stubFor(get(urlEqualTo("/")).willReturn(aResponse().withBody("body is ignored in this test")));
+        getWireMockServer()
```
johnament: Use the static method here.

PR #51, opened 2017-12-05
Make HTML5 rendering include attributes.

johnament:


OndrejM:
For those interested what this PR does, it adds version, status and release information to the HTML output as on the screenshot below. This makes it in line with the PDF version.

image

johnament:
I have one more change (still testing) to make the javadocs deploy as well as a part of the spec release.

Issue #36, opened 2017-11-20
Deployment Exceptions


andymc12:
Need to handle cases where the application code is in error - these should be documented in the spec and tested in the TCK.  Here are a few of the exceptions we should cover (please add more as we think of them):

* Multiple HTTP method annotations on the same interface method (i.e. `@GET @PUT public Object someMethod()`
* Invalid URI templating (i.e. `@GET @Path("/blah") public Object someMethod(@PathParam("{someParam}")` - in this case there is no path parameter defined
* No `@Path` annotation on the passed-in interface

johnament:
Note - we can only do deployment exceptions for CDI managed clients.  We should define an exception to be thrown when you build a client via the builder as well.

andymc12:
Per discussion yesterday, we decided that we do want to support interfaces without an `@Path` annotation (bullet three above).  That means that we are primarily concerned with the following conditions that would define an invalid interface:
* Multiple HTTP method annotations on the same method
* Invalid URI templating (i.e. PathParams parameters without a valid path template in the URI)

From reading the CDI spec on the different exception types, I think that for the CDI case, these should result in `javax.enterprise.inject.spi.DefinitionException`.  For the case where an invalid interface is passed into the RestClientBuilder's build method, the implementation should throw an IllegalArgumentException.

Emily-Jiang:
I suggest you create a separate exception such as RestClientDefinitionException. For CDI managed beans, this exception will be wrapped under DefinitionException. For builder pattern client, it will be bubbled up without being wrapped or wrapped by JAX-RS.

OndrejM:
I think this is rather a nice-to-have feature than something essential. It shouldn't block releasing 1.0 version. If it doesn't make it into 1.0 it can specified in future versions. I don't see any risk in implementations validating this in different ways or not at all as no business logic should rely on exceptions being triggered.

johnament:
I think the scope covered in the PR is enough.  The big question is whether or not the validations it lists can be covered in an impl.

andymc12:
Another condition we might want to check is when a rest client interface declares a sub-resource locator method that returns a class instead of an interface.

PR #52, opened 2017-12-06
Explain how invalid rest client interfaces are handled (#36)

andymc12:
Includes spec, API and TCK updates.

Resolves issue #36

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientDefinitionException.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientDefinitionException.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/RestClientDefinitionException.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * This exception is thrown when the MicroProfile Rest Client implementation
+ * attempts to build a client using an invalid interface.  Interfaces are
+ * considered invalid (1) if it contains a method with more than one HTTP method
+ * annotations or (2) if the combined type-level/method-level URI path contains
+ * an unresolved URI template or (3) if an interface method contains a
+ * <code>@PathParam</code> annotation that refers to a URI template that is not
+ * defined in any <code>@Path</code> annotations on the method or interface.
+ */
+public class RestClientDefinitionException extends RuntimeException {
```
johnament: Can you add the 4 normal constructors?

andymc12: Good catch - I've added them to this PR.

Issue #58, opened 2017-12-16
Various TCK bugs


johnament:
The TCK tests will fail, atleast not behave consistently, because the built in providers will tend to have the same default priority as a JAX-RS runtime's providers.

This means that our message body readers/writers will have the same priority as the default ones.

johnament:
A couple more issues:

- `AdditionalRegistrationTest` has the contracts backwards.
- `FeatureRegistrationTest` doesn't include necessary MP Config entries

PR #59, opened 2017-12-16
More TCK fixes (#58)

johnament:
Could you also take a look @andymc12


johnament:
@andymc12 correct, I'm seeing something weird within CXF.  Maybe we should discuss it over there?

PR #69, opened 2018-01-18
Challenged TCK tests

heiko-braun:
I think some of the TCK tests are incorrect. Maybe someone can look over it and provide feedback if my assumptions are correct?

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
@@ -94,7 +94,7 @@ public class InvokeWithBuiltProvidersTest extends WiremockArquillianTest {
 
         InterfaceWithoutProvidersDefined api = createClient();
 
-        Response response = api.executePut(id, inputBody);
+        Response response = api.executePut(expectedId, inputBody);
 
         String body = response.readEntity(String.class);
 
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
@@ -94,7 +94,7 @@ public class InvokeWithBuiltProvidersTest extends WiremockArquillianTest {
 
         InterfaceWithoutProvidersDefined api = createClient();
 
-        Response response = api.executePut(expectedId, inputBody);
+        Response response = api.executePut(id, inputBody);
 
         String body = response.readEntity(String.class);
 
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
@@ -102,7 +102,9 @@ public class InvokeWithJsonPProviderTest extends WiremockArquillianTest{
         reset();
         stubFor(get(urlEqualTo("/"))
             .willReturn(aResponse()
-                .withBody("[{\"key\": \"value\"}, {\"key\": \"anotherValue\"}]")));
+                .withHeader("Content-Type", "application/json")
+                .withBody("[{\"key\": \"value\"}, {\"key\": \"anotherValue\"}]"))
+                    );
         JsonArray jsonArray = client.get();
         assertEquals(jsonArray.size(), 2, "Expected 2 values in the array for client "+clientType);
         List<JsonObject> jsonObjects = jsonArray.getValuesAs(JsonObject.class);
@@ -120,6 +122,7 @@ public class InvokeWithJsonPProviderTest extends WiremockArquillianTest{
         reset();
         stubFor(get(urlEqualTo("/id"))
             .willReturn(aResponse()
+                .withHeader("Content-Type", "application/json")
                 .withBody("{\"key\": \"value\"}")));
         JsonObject jsonObject = client.get("id");
         assertEquals(jsonObject.keySet().size(), 1, "There should only be one key in object for client "+clientType);
@@ -129,7 +132,10 @@ public class InvokeWithJsonPProviderTest extends WiremockArquillianTest{
 
     private void testPost(JsonPClient client, String clientType) {
         reset();
-        stubFor(post(urlEqualTo("/")).willReturn(aResponse().withStatus(200)));
+        stubFor(post(urlEqualTo("/"))
+                    .willReturn(aResponse()
+                                .withHeader("Content-Type", "application/json")
+                                 .withStatus(200)));
 
         JsonObject jsonObject = Json.createObjectBuilder().add("someKey", "newValue").build();
         String jsonObjectAsString = jsonObject.toString();
@@ -142,7 +148,10 @@ public class InvokeWithJsonPProviderTest extends WiremockArquillianTest{
 
     private void testPut(JsonPClient client, String clientType) {
         reset();
-        stubFor(put(urlEqualTo("/id")).willReturn(aResponse().withStatus(200).withBody("{\"someOtherKey\":\"newValue\"}")));
+        stubFor(put(urlEqualTo("/id"))
+                    .willReturn(aResponse()
+                                .withHeader("Content-Type", "application/json")
+                                .withStatus(200).withBody("{\"someOtherKey\":\"newValue\"}")));
 
         JsonObject jsonObject = Json.createObjectBuilder().add("someKey", "newValue").build();
         String jsonObjectAsString = jsonObject.toString();
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
@@ -97,7 +97,7 @@ public class InvokeWithRegisteredProvidersTest extends WiremockArquillianTest {
             .baseUrl(getServerURL())
             .build(InterfaceWithProvidersDefined.class);
 
-        Response response = api.executePut(id, inputBody);
+        Response response = api.executePut(expectedId, inputBody);
 
         String body = response.readEntity(String.class);
 
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
@@ -97,7 +97,7 @@ public class InvokeWithRegisteredProvidersTest extends WiremockArquillianTest {
             .baseUrl(getServerURL())
             .build(InterfaceWithProvidersDefined.class);
 
-        Response response = api.executePut(expectedId, inputBody);
+        Response response = api.executePut(id, inputBody);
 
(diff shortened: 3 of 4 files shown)


johnament:
While reading the diff, it's not clear what you're challenging, or what you feel should change.  Could you please elaborate a bit more?

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
```diff
@@ -94,7 +94,7 @@ public void testInvokesPutOperationWithRegisteredProviders() throws Exception {
 
         InterfaceWithoutProvidersDefined api = createClient();
 
-        Response response = api.executePut(id, inputBody);
+        Response response = api.executePut(expectedId, inputBody);
```
johnament: This test checks that the providers are invoked, not the server behavior.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
```diff
@@ -97,7 +97,7 @@ public void testInvokesPutOperationWithAnnotatedProviders() throws Exception {
             .baseUrl(getServerURL())
             .build(InterfaceWithProvidersDefined.class);
 
-        Response response = api.executePut(id, inputBody);
+        Response response = api.executePut(expectedId, inputBody);
```
johnament: This test checks that the providers are invoked, not the server behavior.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/CDIInvokeWithRegisteredProvidersTest.java
```diff
@@ -108,7 +108,7 @@ public void testInvokesPutOperation() throws Exception {
             .willReturn(aResponse()
                 .withBody(outputBody)));
 
-        Response response = api.executePut(id, inputBody);
+        Response response = api.executePut(expectedId, inputBody);
```
johnament: This test checks that the providers are invoked, not the server behavior.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
```diff
@@ -142,7 +148,10 @@ private void testPost(JsonPClient client, String clientType) {
 
     private void testPut(JsonPClient client, String clientType) {
         reset();
-        stubFor(put(urlEqualTo("/id")).willReturn(aResponse().withStatus(200).withBody("{\"someOtherKey\":\"newValue\"}")));
+        stubFor(put(urlEqualTo("/id"))
```
johnament: `Content-Type` should be immaterial to this test.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
```diff
@@ -129,7 +132,10 @@ private void testGetSingle(JsonPClient client, String clientType) {
 
     private void testPost(JsonPClient client, String clientType) {
         reset();
-        stubFor(post(urlEqualTo("/")).willReturn(aResponse().withStatus(200)));
+        stubFor(post(urlEqualTo("/"))
```
johnament: `Content-Type` should be immaterial to this test.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
```diff
@@ -102,7 +102,9 @@ private void testGet(JsonPClient client, String clientType) {
         reset();
         stubFor(get(urlEqualTo("/"))
             .willReturn(aResponse()
-                .withBody("[{\"key\": \"value\"}, {\"key\": \"anotherValue\"}]")));
+                .withHeader("Content-Type", "application/json")
```
johnament: `Content-Type` should be immaterial to this test.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
heiko-braun: The wiremock setup in this test registers `/expectedId`, hence any invocation to `/id` fails with a `404` .

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
heiko-braun: I don't know if this anchored in the HTTP or JAX-RS specs, but without a valid `Content-type` header my underlying JAX-RS implementation does not pick up the JSON  reader.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
heiko-braun: See my comment above.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithRegisteredProvidersTest.java
heiko-braun: See my comment above

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/CDIInvokeWithRegisteredProvidersTest.java
heiko-braun: See my comment above

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
johnament: Yeah, I figured that's what the issue was.  I'm not against making the change, but your impl should pass without this.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
johnament: Ah ha.  You should be activating the JSON Reader simply because the method returns a `JsonArray`

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
johnament: Actually, now that I'm thinking about it, the CXF impl should be seeing the 404 getting thrown as an exception as well (assuming that's what you're seeing in your impl)

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
johnament: BTW, the spec mentions "OR" between the produces/consumes/method return values.  [Take a look.](https://github.com/eclipse/microprofile-rest-client/blob/master/spec/src/main/asciidoc/providers.asciidoc#json-p-provider)

heiko-braun: I don't want lengthy arguments about this, but generally speaking content-type should be present in HTTP. This doesn't change the nature of the test, but rather refines it, so why not simply include it? 

The impl we leverage (RESTEasy) chokes on it, and this would be a simple fix, which I think is fully aligned with the test assertion and the HTTP spec.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithBuiltProvidersTest.java
heiko-braun: I see, thanks for the clarification. It seems I missed that subtle involvement of the param  converter

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
johnament: @heiko-braun i understand your point.  I would recommend that we change the `Content Type` to be `text/plain` to ensure that it's being activated based on the object in use, not the media of the server response.

heiko-braun: I believe I get the idea, however in RESTEasy this leads to `Unable to find a MessageBodyReader of content-type text/plain and type interface javax.json.JsonObject`

andymc12:
FWIW, I agree that the Content-type should not be required, but I'm fine with adding it to the response - I don't think it hurts the test cases.

The "id" vs "expectedId" change (as it is now) should not go in, since part of the test is to verify that the ParamConverter works.  I'd be fine if we just leave those tests alone - or (preferred) make it clearer that id is expected to be converted (with a comment or renaming "id" to "preConvertedId" or something like that).

Thanks, Andy

heiko-braun:
Yes, I agree. The ParamConverter test can stay as it is. I'll provide an update to this PR next week that reduces it to just the JsonPProviderTest. Does that sound resonable?

heiko-braun:
@johnament @andymc12 I've restored the tests that rely on ParamConverter. The change set is reduced to just the `InvokeWithJsonPProviderTest.java`

heiko-braun:
@andymc12 I've opened a discussion about the `ParamConverter`: https://github.com/eclipse/microprofile-rest-client/issues/74

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java
johnament: FWIW, I'm fine with the final resolution of this however I have an application running today that uses resteasy, receives content type of text/html and deserializes it as JSON.  It's all about fixing the produces/consumes on the MBR/MBW.

johnament: Ok, so I ended up digging into the spec a bit to understand what in the world resteasy is doing.

I found this commit: https://github.com/resteasy/Resteasy/commit/fb33d3c1bac91bdb1d686008bb679fae7044820d

It references sections of the spec.  So I opened up the spec, and found these notes:

> • If the method is annotated with @Produces, set P = {V (method)} where V (t) represents the
values of @Produces on the specified target t.
> • Else if the class is annotated with @Produces, set P = {V (class)}.
> • Else set P = {V (writers)} where ‘writers’ is the set of MessageBodyWriter that support the
class of the returned entity object.

Since the interface doesn't define any [https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/JsonPClient.java](produces/consumes), it would appear that resteasy is doing the wrong thing and looking at the content type to determine the match.  Based on the spec content I copied above, I would expect that it falls into bullet 3 and uses the class of the returned entity to be invoked (noting that in fact we should be using MBR instead of MBW in the type safe client).

If you like, I can create a ticket against resteasy.

Issue #76, opened 2018-02-20
Prefer URI to URL


hwellmann:
Given that JAX-RS normally uses `URI` instead of `URL`, `RestClientBuilder` should have a method

    RestClientBuilder baseUri(URI uri);

in addition to

    RestClientBuilder baseUrl(URL url);

and there may be good reasons to deprecate the latter.

andymc12:
Resolved by #78

PR #78, opened 2018-02-21
[#76] Add new baseUri method

andymc12:
- Adds new `baseUri` method
- Deprecated `baseUrl` method
- Changing existing tests to call `baseUri`
- New test for ensuring last call to either `baseUri` or `baseUrl` "wins"
- New test for ensuring `RestClientBuilder` throws an `IllegalStateException` if no address was specified.
- Updated release notes to indicate changes

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -16,6 +16,7 @@
 package org.eclipse.microprofile.rest.client;
 
 import javax.ws.rs.core.Configurable;
+import java.net.URI;
 import java.net.URL;
 import org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver;
 
@@ -50,10 +51,35 @@ public interface RestClientBuilder extends Configurable<RestClientBuilder> {
      * <code>http://my-service:8080/service/api</code> in addition to any
      * <code>@Path</code> annotations included on the method.
      *
+     * The {@link #baseUri(URI)} method should be preferred over this method.
+     *
+     * Subsequent calls to this method will replace the previously specified
+     * baseUri/baseUrl.
+     *
      * @param url the base Url for the service.
      * @return the current builder with the baseUrl set
+     * @deprecated use {@link #baseUri(java.net.URI)} instead.
      */
+    @Deprecated
     RestClientBuilder baseUrl(URL url);
+
+    /**
+     * Specifies the base URI to be used when making requests. Assuming that the
+     * interface has a <code>@Path("/api")</code> at the interface level and a
+     * <code>uri</code> is given with
+     * <code>http://my-service:8080/service</code> then all REST calls will be
+     * invoked with a <code>uri</code> of
+     * <code>http://my-service:8080/service/api</code> in addition to any
+     * <code>@Path</code> annotations included on the method.
+     *
+     * Subsequent calls to this method will replace the previously specified
+     * baseUri/baseUrl.
+     *
+     * @param uri the base URI for the service.
+     * @return the current builder with the baseUri set
+     * @since 1.1
+     */
+    RestClientBuilder baseUri(URI uri);
 
     /**
      * Based on the configured RestClientBuilder, creates a new instance of the
@@ -63,8 +89,8 @@ public interface RestClientBuilder extends Configurable<RestClientBuilder> {
      * @param <T> the type of the interface
      * @return a new instance of an implementation of this REST interface that
      * @throws IllegalStateException if not all pre-requisites are satisfied for
-     * the builder, this exception may get thrown. For instance, if a URL has
-     * not been set.
+     * the builder, this exception may get thrown. For instance, if the base
+     * URI/URL has not been set.
      * @throws RestClientDefinitionException if the passed-in interface class is
      * invalid.
      */
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -16,6 +16,7 @@
 package org.eclipse.microprofile.rest.client;
 
 import javax.ws.rs.core.Configurable;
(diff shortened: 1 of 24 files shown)


andymc12:
Resolves issue #76

andymc12:
Hi @Emily-Jiang - yes, it is still tested in the `ProvidesRestClientBuilderTest` test.

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -50,11 +51,36 @@ static RestClientBuilder newBuilder() {
      * <code>http://my-service:8080/service/api</code> in addition to any
      * <code>@Path</code> annotations included on the method.
      *
+     * The {@link #baseUri(URI)} method should be preferred over this method.
```
johnament: Saying that `baseUri` is preferred is misleading.  Are you saying that using the following chain, the value of `baseUri` is used?

`newBuilder().baseUri(someUri).baseUrl(someUrl).proxy(someClass)`

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -50,11 +51,36 @@ static RestClientBuilder newBuilder() {
...
+     *
+     * Subsequent calls to this method will replace the previously specified
+     * baseUri/baseUrl.
+     *
      * @param url the base Url for the service.
      * @return the current builder with the baseUrl set
+     * @deprecated use {@link #baseUri(java.net.URI)} instead.
      */
+    @Deprecated
```
johnament: I don't know that I agree with deprecating it.  I agree with adding support for URI objects, but it seems a bit extreme to say that this is now deprecated.  FWIW, I have 0 recollection on how we ended up on URL; I think I had assumed that `URL` was supported across the board.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -74,7 +74,7 @@ public interface MyServiceClient {
 
 The values of the following properties will be provided via MicroProfile Config:
 
-- `com.mycompany.remoteServices.MyServiceClient/mp-rest/url`: The base URL to use for this service, the equivalent of the `baseUrl` method.  This property is considered required, however implementations may have other ways to define these URLs.
+- `com.mycompany.remoteServices.MyServiceClient/mp-rest/url`: The base URL to use for this service, the equivalent of the `baseUri` method.  This property is considered required, however implementations may have other ways to define these URLs.
```
johnament: should we also support `*/mp-rest/uri`?

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -50,11 +51,36 @@ static RestClientBuilder newBuilder() {
...
+     * <code>@Path</code> annotations included on the method.
+     *
+     * Subsequent calls to this method will replace the previously specified
+     * baseUri/baseUrl.
+     *
+     * @param uri the base URI for the service.
+     * @return the current builder with the baseUri set
+     * @since 1.1
+     */
+    RestClientBuilder baseUri(URI uri);
```
johnament: To avoid API breakage, can this be a default method that performs `return baseUrl(uri.toURL())`?

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
andymc12: Agreed - it is not good wording.  In that case, no - the last call (to either baseUri or baseUrl) wins.  I've got a TCK test that confirms that, but I will work on the wording to indicate that baseUri is only "preferred" in that URIs are preferred to URLs - in the google groups, it was mentioned that URLs have some blocking methods where some methods will result in network activity - see http://blog.markfeeney.com/2010/11/java-uri-vs-url.html for one instance.

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
andymc12: Right, I don't remember discussing it either.  I saw that other proxy-based REST client implementations also use URLs, so maybe that is why.  The argument to deprecate is driven by the network activity issues mentioned above.  There is also the issue that constructing URLs requires that the user catch a MalformedURLException.  Users could use `URI.create(someUrl)` to create a URI without needing to catch any exceptions.

spec/src/main/asciidoc/cdi.asciidoc
andymc12: Good question... we probably should, and then deprecate the `*/mp-rest/url` config property (assuming the consensus is to deprecate the `baseUrl` method).  Any strong opinions?

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
andymc12: Yes, great suggestion!

andymc12:
In this latest commit, I removed the controversial "is preferred" line from the javadoc and I changed the `baseUri` to be a default method that calls `baseUrl(uri.toURL())` for backward compatibility.

I'd like more feedback on:
1) Do we really want to deprecate `baseUrl`?  I'm leaning toward deprecating, but could be persuaded.
2) Should we add a new `*/mp-rest/uri` config property?  I don't have a strong opinion on this either way, but I think it is a good question.

Thanks, Andy

Emily-Jiang:
I think we just support both url and uri equally. Whatever you did for url should do the same for uri if applicable.

andymc12:
For 1, I'll plan to _not_ deprecate the `baseUrl` method.
For 2, I think we should add the new config property until we have an MP Config converter that handles URIs - I added this requirement on MP Config in Issue [322](https://github.com/eclipse/microprofile-config/issues/322).

Emily-Jiang:
@andymc12 the current MP Config should be able to directly treat URI as an implicit converter. I don't think you are blocked.

Issue #79, opened 2018-02-23
Enable support for global provider registration


andymc12:
Issue #35 expose the need for a global or system-wide registration of providers.  In the OpenTracing example, the OpenTracing implementation could easily add a ClientRequestFilter that would start the tracing, but then user code would be responsible for registering the filter.  Instead, there should be a mechanism that allows automatic registration of providers to be used for all rest clients.

I'd propose using the ServiceLoader mechanism to search the classpath for instances of an interface that would register these providers.  In this way, OpenTracing could add an implementation of that interface to register their filter.  Likewise, other features (Fault Tolerance, Metrics, or even user-specific features) could register their own providers.

andymc12:
Resolved by #80

PR #80, opened 2018-02-23
[Issue #79] RestClientBuilderListener for global provider registration

andymc12:
This would resolve issue #79.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderListener.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderListener.java
@@ -1,0 +1,45 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.rest.client.spi;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+
+/**
+ * Implementations of this interface will be notified when new RestClientBuilder
+ * instances are being constructed.  This will allow implementations to register
+ * providers on the RestClientBuilder, and is intended for global providers.
+ * For example, a MicroProfile OpenTracing implementation might want to register
+ * a ClientRequestFilter to initiate tracing.
+ *
+ * In order for the RestClientBuilder to call implementations of this interface,
+ * the implementation must be specified such that a ServiceLoader can find it -
+ * i.e. it must be specified in the <code>
+ * META-INF/services/org.eclipse.microprofile.rest.client.spi.RestClientBuilderListener
+ * </code> file in an archive on the current thread's context classloader's
+ * class path.
+ *
+ * Note that the <code>onNewBuilder</code> method will be called when the
+ * RestClientBuilder is constructed, not when it's <code>build</code> method is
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/spi/ProviderRegistrar.java
```diff
@@ -0,0 +1,45 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2017-2018 Contributors to the Eclipse Foundation
```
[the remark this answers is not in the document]
andymc12: Will fix - good catch.

andymc12:
@Emily-Jiang I chose the ServiceLoader pattern so that user apps could use this approach too. The use case for that would be similar to OpenTracing: suppose a user wants to log all outbound MP Rest Client calls - but maybe not using MP OpenTracing.  They could do their own logging in a filter in a similar fashion.  There are other use cases - like adding a custom header to all requests coming from a particular client endpoint, or registering a default set of ResponseExceptionMappers that are common across an entire application (thus avoiding the more verbose option of explicitly registering each mapper with each interface).

Could you send me a link to more info on the MP listener/registration pattern?  If it would work for users too, then it might be preferable to the ServiceLoader pattern that I have implemented here.

johnament:
There is no consistent listener/registration pattern.  I know something was just added recently to ConfigJSR when you add a listener, but none of the actual MP specs use it.

api/src/main/java/org/eclipse/microprofile/rest/client/spi/ProviderRegistrar.java
```diff
@@ -0,0 +1,45 @@
...
+ * </code> file in an archive on the current thread's context classloader's
+ * class path.
+ *
+ * Note that the <code>newBuilder</code> method will be called when the
+ * RestClientBuilder is constructed, not when it's <code>build</code> method is
+ * invoked.  This allows the caller to override global providers if they desire.
+ */
+public interface ProviderRegistrar {
+
+    void newBuilder(RestClientBuilder builder);
```
johnament: Though I do see what Emily means.  the method `newBuilder` is a life cycle method when a new builder is created.   There's a few other things we could do:

- Create an `AutoRegisteredClientFeature` interface which extends `Feature`, used on bootstrap to load features that should always be applied.
- Rename this method from `newBuilder` to `onNewBuilder` or even `customize` to better explain its purpose.

spec/src/main/asciidoc/providers.asciidoc
```diff
@@ -90,6 +90,8 @@ Any methods that read the response body as a stream must ensure that they reset
 
 In addition to defining providers via the client definition, interfaces may use the `@RegisterProvider` annotation to define classes to be registered as providers in addition to providers registered via the `RestClientBuilder`
 
+Providers may also be registered by implementing the `ProviderRegistrar` interface.  This interface is intended as an SPI to allow global provider registration.  The implementation of this interface must be specified in a META-INF/services/org.eclipse.microprofile.rest.client.spi.ProviderRegistrar file, following the service loader pattern.
```
johnament: The `META-INF` file should be listed as a code.

api/src/main/java/org/eclipse/microprofile/rest/client/spi/ProviderRegistrar.java
andymc12: Makes sense.  Personally, I like `onNewBuilder` the best.  Will change.

api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderListener.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.rest.client.spi;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+
+/**
+ * Implementations of this interface will be notified when new RestClientBuilder
+ * instances are being contructed.  This will allow implementations to register
```
Emily-Jiang: constructed

Emily-Jiang:
@johnament was right. We discussed this in configjsr. @struberg recently ported this back to MP Config. https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/Config.java I understand a bit more of your use case. Maybe service loader pattern fits more as it is automatic.

andymc12:
Thanks @Emily-Jiang - I updated the version to 1.1 in the package-info.java file and corrected the typo.  Thanks for catching these issues.

andymc12:
Merged with other changes - and will merge to master per John's review in gitter channel.

PR #83, opened 2018-02-26
Add @Stereotype to @RegisterRestClient

andymc12:
This will ensure that the `@RegisterRestClient` is a bean-defining annotation.


johnament:
If all you're trying to do is make it a bean defining annotation, `@Stereotype` seems like an odd choice.  Since the default scope is `@Dependent` that would make more sense here, in addition to `@Stereotype`

andymc12:
That makes sense @johnament - I will add the `@Dependent` annotation as well.  Thanks for the quick review!

Emily-Jiang:
@johnament strictly speaking, when defining Stereotype with specifying any scope, Depended is used. Maybe adding it explicitly is better. However, having it absent is also fine as it will be defaulted.

Issue #10, opened 2017-10-09
Async client invocations


johnament:
Define how to use async invocations

andymc12:
Do we think that this is still appropriate for MP Rest Client 1.0?  My thought for this was something like:
```
@Path("/root")
public interface MyClient {
    @GET
    Future<MyResponseObject> get1();

    // and/or

    @GET
    CompletionStage<MyResponseObject> get2();
}
...
MyClient mc = RestClientBuilder.newBuilder()
                            .executorService(myExecutorService) // <-- new method to builder - based on JAX-RS 2.1
                            .baseUrl(myURL)
                            .build(MyClient.class);
```

rquinio:
I was just proposing something like this CXF mailing list.
Attempting to translate JAX-WS async proxy callback approach using JAX-RS types, how about:

```
@Path("/root")
public interface MyClient {
    @GET
    Future<?> get1(InvocationCallback<MyResponseObject> callback);
}
```
i.e the Future<?> object returned can only be used to test whether or not a response has arrived yet, while the value of the response is only made available to the callback object.
Since JAX-RS async WebClients would already use javax.ws.rs.client.InvocationCallback, this could possibly ease migration to type-safe proxies ?

kenfinnigan:
In WF Swarm we offer a [ServiceClient](https://github.com/wildfly-swarm/wildfly-swarm/blob/master/client-apis/jaxrs-client-api/src/main/java/org/wildfly/swarm/client/jaxrs/ServiceClient.java) interface to provide async methods.

So we can have an interface definition such as:
```
@Path("/")
public interface TimeService extends ServiceClient<TimeService> {
    @GET
    @Produces(MediaType.TEXT_PLAIN)
    String getTime();
}
```

Which can then be used in an async fashion like
```
@GET
@Path("/async")
public void getMessageAsync(@Suspended final AsyncResponse asyncResponse) throws Exception {
    timeService.exec(timeService::getTime,
                     s -> asyncResponse.resume(s),
                     asyncResponse::resume);
}
```

hutchig:
I was interested in how the microprofile-rest-client could be used in a "reactive manner".
cf https://spring.io/guides/gs/reactive-rest-service/ 
I think it would be great if the major Jakarta specs (via microprofile of course :-) ) adopted reactive streams (https://github.com/eclipse/microprofile-reactive) so they could be plugged together with functions compostion and also get backpressure.

hutchig:
See also #105

Issue #35, opened 2017-11-20
Allow to start traces in the client


pilhuhn:
When MP-OpenTracing is in the server, it should be possible to start a trace already in the client, so that the network hop between client and server is also recorded as a span.

cc @pavolloffay

pavolloffay:
From tracing perspective we need:
1. a way to inject tracing context into the headers
2. start a span for outbound requests
3. wire server span with client span so the client span is a child of the server span (if threre is any). In jax-rs there are some problems with async requests see https://github.com/opentracing-contrib/java-jaxrs#async

johnament:
Couldn't this be handled with a `ClientRequestFilter`?

johnament:
The equivalent of this is https://github.com/eclipse/microprofile-opentracing/issues/14 which will be updated to indice that any Client created will automatically have traces enabled.

We should update the MP Rest Client spec to also indicate this.

But I'm not sure the best way to add a TCK test for this.  Needs more thought.

johnament:
Based on the conversation from yesterday, we're not going to try to tackle this in 1.0.  We are likely going to push for a platform level spec that describes how some of these specs work together.

OndrejM:
As I expressed on the mailing list, I don't like specifying this cross-spec functionality only on the umbrella level. To me, REST client spec is clearly the dominant spec for this feature and should specify how REST clients behave if opentracing is supported, including how to disable this behavior if not desired. Then it's clear how to integrate REST client and opentracing together even without implementing all MP specs. 

This is the same pattern as JTA specifying `@Transactional` annotation when used in a CDI container. It helps implementation vendors to test components separately as they are implemneted and also helps to test componentized MP implementations like WildFlySwarm fractions and ensure that every combination of components works as expected.

OndrejM:
Also, I don't see how we would effectively specify this functionality in the umbrella spec. Until now, it's been a very lightweight umbrella wrapping a set of specs together with minimal effort and very shortly after they reached a final version. There's not much time to write integration tests between individual specs are final and the umbrella spec is final. The tests would need to be being written in parallel to developing individual specs. And in that case, it adds one more repository to track and update - instead of just 2 (opentracing and rest-client) there would be additional umbrella repo. There will be cases where we can't avoid it because they wouldn't fit to any individual TCK but the case described in this issue can easily fit into the REST client TCK with the guidance of people involved in the opentracing spec.

derekm:
[...]

PR #85, opened 2018-03-07
Issue #10: Add async support

andymc12:
First pass at async support - resolves Issue #10.

This adds:
1) AsyncInvocationInterceptorFactory/AsyncInvocationInterceptor
2) General discussion of what constitutes an async method
3) New executorService(...) method for RestClientBuilder
4) Basic TCKs


spec/src/main/asciidoc/async.asciidoc
```diff
@@ -0,0 +1,86 @@
...
+There may be cases where it is necessary for client application code or runtime components to be notified when control of the client request/response has flowed from one thread to another.
+This can be accomplished by registering an implementation of the AsyncInvocationInterceptorFactory provider interface.
+MP Rest Client implementations must invoke the `newInterceptor` method of each registered factory provider prior to switching thread of execution on async method requests.
+That method will return an instance of `AsyncInvocationInterceptor` - the MP Rest Client implementation must then invoke the `pre` method while still executing on the thread that invoked the async method.
+After swapping threads, but before invoking further providers or returning control back to the async method caller, the MP Rest Client implementation must invoke the `post` method on the new async thread that will complete the request/response.
+
+The following example shows how the AsyncInvocationInterceptorFactory provider and associated AsyncInvocationInterceptor interface could be used to propagate a `ThreadLocal` value from the originating thread to async thread:
+[source, java]
+----
+@Provider
```
aguibert: The usage of the JAX-RS `@Provider` annotation here seems inconsistent with the usage of `@RegisterProvider` throughout the rest of MP Rest Client APIs.  If I'm understanding correctly that AIIF and AII are meant to be implemented by users of MP Rest Client (as opposed to implementers of MP Rest Client), it would make more sense for the user to define `@RegisterProvider(AsyncInvocationInterceptorFactory.class)` on the JAX-RS service interface.  If this API is only meant to be used by implementers of MP Rest Client, you can ignore this comment.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/InvocationCallbackClientAsync.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.eclipse.microprofile.rest.client.tck.interfaces;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.client.InvocationCallback;
+
+@Path("/string")
+public interface InvocationCallbackClientAsync {
```
aguibert: Isn't this missing `@Dependent` and `@RegisterRestClient`?

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/SimpleGetApiAsync.java
```diff
@@ -0,0 +1,32 @@
...
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+@Path("/")
+@RegisterRestClient
+public interface SimpleGetApiAsync {
```
aguibert: Is this missing `@Dependant`?

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/StringResponseClientAsync.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.eclipse.microprofile.rest.client.tck.interfaces;
+
+import java.util.concurrent.Future;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+@Path("/string")
+public interface StringResponseClientAsync {
```
aguibert: same comment about missing annotations

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+ * implementing this interface to avoid race conditions and deadlocks.
+ */
+public interface AsyncInvocationInterceptor {
+
+    /**
+     * This method will be invoked by the MP Rest Client runtime on the "main"
+     * thread (i.e. the thread calling the async Rest Client interface method)
+     * prior to returning control to the calling method.
+     */
+    void pre();
```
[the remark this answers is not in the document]
andymc12: I'm not opposed to changing the method names, but the meaning should convey that we are swapping threads which isn't really an invocation.  The `pre` method is really `justAboutToContinueThisRESTfulRequestOnAnotherThread" and the `post` method is really `nowDoingWorkOnTheNewAsyncThread`.  I definitely think there is a better name for these methods, but I haven't found it yet.  Please let me know if you've got other ideas for it though.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+ * method on the main thread prior to returning execution back to the calling
+ * method of the client interface.  The runtime will invoke the
+ * <code>post</code> method on the secondary thread before the client request is
+ * sent.  The <code>pre</code> method should always be invoked before the
+ * <code>post</code> method is invoked, but due to the nature of multithreading,
+ * it is possible that <code>post</code> method may be invoked before the
+ * <code>pre</code> method has completed.  Care should be taken when
+ * implementing this interface to avoid race conditions and deadlocks.
+ */
+public interface AsyncInvocationInterceptor {
```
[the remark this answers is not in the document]
andymc12: Agreed - will add a brief reference here.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptorFactory.java
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.microprofile.rest.client.ext;
+
+/**
+ * This is a provider interface intended for intercepting asynchronous method
+ * invocations.  Registered implementations of this interface will be invoked
+ * to build a new <code>AsyncInvocationInterceptor</code>.  This interceptor
+ * would then be invoked before and after swapping threads for an asynchronous
+ * client invocation.
+ *
+ * The priority of this provider determines when the factory's
```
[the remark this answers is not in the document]
andymc12: Good idea - will be more explicit.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+     * prior to returning control to the calling method.
+     */
+    void pre();
+
+    /**
+     * This method will be invoked by the MP Rest Client runtime on the "async"
+     * thread (i.e. the thread used to actually invoke the remote service and
+     * wait for the response) prior to sending the request.
+     */
+    void post();
```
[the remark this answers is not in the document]
andymc12: I'm leaning towards leaving them as non-default. Most of the use cases for these interceptors involve implementing both sides - pre and post.  Do you have another use case in mind that might only use one or the other?

spec/src/main/asciidoc/async.asciidoc
andymc12: Agreed that the `@Provider` annotation is out of place.  The `AsyncInvocationInterceptorFactory` pattern is primarily intended for "add-on" components to the MP Rest Client - like OpenTracing - possibly Metrics, etc., but there is no reason that a user _couldn't_ use it.  I'll plan to remove the `@Provider` and add `@RegisterProvider(...)` to make it more clear.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/InvocationCallbackClientAsync.java
andymc12: No, those annotations are only required when using CDI.  For these test cases, I'm using the `RestClientBuilder` - if we wanted to use this same client interface in a test that involved CDI, then we would definitely need to add `@RegisterRestClient` - as for `@Dependent`, I'm hoping that that won't be necessary (even in the CDI case) after PR #83 is merged.

This answer should also apply to the questions in `SimpleGetApiAsync.java` (which is used in a CDI case) and `StringResponseClientAsync.java`.

spec/src/main/asciidoc/async.asciidoc
```diff
@@ -0,0 +1,86 @@
...
+[[restasync]]
+== MicroProfile Rest Client Asynchronous Support
+
+It is possible for Rest Client interface methods to be declared asynchronous.  This allows the thread invoking the interface method to proceed while the RESTful request occurs on another thread.
+
+=== Asynchronous Methods
+
+A method is considered to be asynchronous if either of the following conditions are true:
+
+- The method's return type is `java.util.concurrent.Future`.
```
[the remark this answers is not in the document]
andymc12: I'd be interested to hear from other parties on this.  Another (possibly simpler) option would be to drop support for `Future` return types, and only perform async on methods with an `InvocationCallback` parameter.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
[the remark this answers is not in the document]
aguibert: oh wow, I had completely misunderstood the meaning of these methods then.  I thought they were invoked before/after the request respectively.  I think these methods definitely need to be renamed in that case. How about `prepareContext/applyContext`?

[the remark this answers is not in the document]
andymc12: I like that a lot better - thanks for the suggestion!

kenfinnigan:
So I've taken a look at the code and I'm wondering if there isn't an alternative approach that makes things a bit simpler from a developer perspective.

As I understand what's proposed, right now a developer needs to implement a Factory to define what is done asynchronously, and then register that Factory on the interface they've defined. In addition any method that what's to be executed asynchronously needs to be modified to return `Future` or pass `InvocationCallback`. To me this seems like lots of boilerplate to need to add to execute a client request in an asynchronous fashion.

Something I developed for WF Swarm a while ago might fit here and simplify the approach. Basically we offer a default interface: 
```
public interface ServiceClient<T> {
    default <U> void exec(Supplier<U> restMethod, Consumer<U> handler, Consumer<Throwable> exceptionHandler) throws Exception {
        CompletableFuture
                .supplyAsync(restMethod, executorService())
                .thenAccept(handler)
                .exceptionally(t -> {
                    exceptionHandler.accept(t);
                    return null;
                });
    }

    default <U> void execAsync(Supplier<U> method, Consumer<U> success, Consumer<Throwable> failure) throws Exception {
        CompletableFuture
                .supplyAsync(method, executorService())
                .thenAccept(success)
                .exceptionally(t -> {
                    failure.accept(t);
                    return null;
                });
    }

    default ManagedExecutorService executorService() throws Exception {
        InitialContext ctx = new InitialContext();
        return (ManagedExecutorService) ctx.lookup("java:jboss/ee/concurrency/executor/default");
    }
}
```
It defines some async execution methods. The idea is that your interface defining the external service then just extends that interface, and the methods on the interface stay as they are.

For instance, if we had an interface defined as:
```
public interface TimeService extends ServiceClient<TimeService> {
    @GET
    @Produces(MediaType.TEXT_PLAIN)
    String getTime();
}
```

It can then be invoked sync:
```
String time = timeService.getTime();
```

Or async:
```
timeService.exec(timeService::getTime,
                     asyncResponse::resume,
                     asyncResponse::resume);
```

Other than defining that the interface extends another interface, you're not changing method definitions between sync or async invocation.

Now I may be not understanding some benefits of the current approach, but this seems to be easier from a developer perspective

andymc12:
Hi @kenfinnigan , thanks for the feedback.  The AsyncInvocationInterceptors (and the associated factories) are not required for async methods - they are only mentioned for cases like OpenTracing that need to track inbound/outbound data and store it in ThreadLocal data structures.  More details are in issue #35 - toward the bottom of the discussion.

As it is currently, the client developer would just need to structure their interface method with either an `InvocationCallback` as a parameter or have the return type be `Future` like:
```
public interface MyAsyncClient {

    @GET
    Future<Response> get();

    @PUT
    Future<SomeObject> put(SomeOtherObject o);

    @POST
    void post(InvocationCallback<AnotherObject> callback, YetAnotherObject requestBody);
}
```

All of these would be automatically treated as async by the MP Rest Client implementation.  I think this should be pretty simple and powerful for most developers.  

What do you think?  Thanks again - Andy

kenfinnigan:
My personal view is that `Future` and `InvocationCallback` are older ways of doing async that aren't always appropriate.

Use of `CompletableFuture` would at least make combining client calls possible, without having to construct one in their own code, which is the approach we took with WF Swarm.

I'd prefer not to implement it this way, but if my opinion is the sole dissenting view, then I won't stop it being included

andymc12:
That makes sense.  When we talked about Async several months ago, we mentioned not using `CompletableFuture` until we re-align with JAX-RS 2.1 (EE8) in MP 2.0.  Technically, `CompletableFuture` is part of Java 8 so we could depend on it now.

I see advantages and disadvantages to both approaches - I'll try to sum them up:
__Future/InvocationCallback:__
Pros:
* aligns with JAX-RS 2.0
* client interfaces are not required to extend any other service client interface
* slightly simpler API
Cons:
* does not work with `CompletableFuture`
* cannot change the `ExecutorService` in use once the client interface impl is built

__ServiceClient:__
Pros:
* Supports reactive-style programming with `CompletableFuture`
* Can change `ExecutorService` after building the client interface impl (assuming we added an `execAsync(..., ExecutorService)` default method - or when using the `CompletableFuture` APIs.
* Sync and Async using the same client interface
Cons:
* Requires client interfaces to extend the service client interface
* Slightly more complex to invoke method asynchronously

Any strong opinions or other thoughts?  @Emily-Jiang @johnament @aguibert

Ladicek:
I totally agree with @kenfinnigan that `Future` and `InvocationCallback` are the old and bad way of doing async. Bad because they don't really allow composition. At the same time, I see the `ServiceClient` approach as too cumbersome and the code in https://github.com/eclipse/microprofile-rest-client/pull/85#issuecomment-375305998 doesn't really support using `CompletableFuture`. (What's in WildFly Swarm is different, though: https://github.com/wildfly-swarm/wildfly-swarm/blob/master/client-apis/jaxrs-client-api/src/main/java/org/wildfly/swarm/client/jaxrs/ServiceClient.java)

What I would personally like to see is the ability to write

```java
public interface MyAsyncClient {
    @GET
    CompletableFuture<Response> get();
}
```

So pretty similar to the original `Future` approach, but with the more modern async API :-) It would be possible to allow `CompletionStage` as well (and maybe any subtype of it, if there's a way to construct it?).

It doesn't support doing sync and async calls with just single method declaration, and I don't see that as a big deal really. Async is just too different from sync that shoehorning them under a single API inevitably leads to compromises. I personally think that the WildFly Swarm approach primarily targets sync calls and punishes anyone who wants to do async. (Full disclosure: I tried to convince @kenfinnigan about this when he was working on the `ServiceClient` in WildFly Swarm, and I only partially succeeded. So this is my second attempt :-D)

One of the advanced usecases is being able to provide the `j.u.c.Executor` that the should be used for the async call. I think it would be possible to mirror the `CompletableFuture` API here (where the async methods have 2 overloads: one using the default executor and the other taking an `Executor` as the last parameter). I believe that `Executor` (and any subtype of it) is never a valid parameter to a REST call, so it would be possible for the interface methods to just declare a (single) parameter of type `Executor` (or subtype) -- and the framework would expect that the caller will just pass any executor they want.

api/src/test/java/org/eclipse/microprofile/rest/client/BuilderImpl2.java
```diff
@@ -32,6 +33,10 @@ public RestClientBuilder baseUri(URI uri) {
         throw new IllegalStateException("not implemented");
     }
 
+    public RestClientBuilder executorService(ExecutorService executor) {
```
johnament: can be in `AbstractBuilder`

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
```diff
@@ -0,0 +1,55 @@
...
+ * multithreading, it is possible that <code>post</code> method may be invoked
+ * before the <code>prepareContext</code> method has completed.  Care should be
+ * taken when implementing this interface to avoid race conditions and
+ * deadlocks.
+ *
+ * Note that the order in which instances of the
+ * <code>AsyncInvocationInterceptor</code> are invoked are determined by the
+ * priority of the <code>AsyncInvocationInterceptorFactory</code> provider.
+ */
+public interface AsyncInvocationInterceptor {
```
johnament: I don't see a lot of use for this interceptor the way its defined right now.  Is there a way it can get access to the underlying resource? or some kind of context object with more information?

it seems like its analogous to `ClientRequestFilter` and `ClientResponseFilter`, but I'm not sure I fully grok it.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptorFactory.java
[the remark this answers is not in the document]
johnament: Also, do we really need a factory? or can the interceptor simply be registered?

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
andymc12: The primary reason for it is to allow implementors to have some flexibility in how they implement async methods (for example, CXF marshalls the request on the main thread while Jersey returns immediately - and does everything on the new thread) but allow for providers (like OpenTracing) to track things in thread locals.  

So an implementation of the factory interface could have `@Context` annotation injection for things like headers, URIInfo, etc. Then the factory could create the actual interceptors with any pertinent data (like OpenTracing info from the main thread, etc.).  Now, if they need to ensure that thread locals get copied over to the new thread, they can do that with the interceptor.

api/src/test/java/org/eclipse/microprofile/rest/client/BuilderImpl2.java
andymc12: I can move it - but it seemed like the `AbstractBuilder` class mainly had methods from `Configurable` while the sub-classes had the `RestClientBuilder` -specific methods.

api/src/main/java/org/eclipse/microprofile/rest/client/ext/AsyncInvocationInterceptor.java
johnament: The OpenTracing part makes sense.  However, the part of the vendors seems a bit confusing.  Is that an implementation detail?  To be clear, the part that's confusing is the use of a `Factory` that creates the interceptor, instead of the developer/framework simply registering the interceptor directly.

andymc12: Ah, sorry for the confusion.

The reason for the Factory is because most use cases of this provider type is per-request, but the standard lifecycle for providers are singleton (per-app).  By using a factory pattern implementors of the factory/interceptors can use either approach.  For example, the OpenTracing use case really needs to be per-request - something like:

```
class MyFactory implements AsyncInvocationInterceptorFactory {
    public AsyncInvocationInterceptor newInterceptor() {
        return new OpenTracingAsyncInterceptor();
    }
}

class OpenTracingAsyncInterceptor implements AsyncInvocationInterceptor {
    private Span span;
    public void prepareContext() {
        span = SomeOtherClass.getSpanFromThreadLocal();
    }

    public void applyContext() {
        SomeOtherClass.setSpanToThreadLocal(span);
    }
}
```

The `span` in this case must be per-request - otherwise wires could get crossed.  But there could be other use cases (like debug logging) where a per-app singleton would be preferred (to avoid excessive garbage, object pooling, etc.)- i.e.:

```
MyLoggingProvider implements AsyncInvocationInterceptorFactory, AsyncInvocationInterceptor {
    static Logger LOG = Logger.getLogger(MyLoggingProvider.class);
 
    @Context
    UriInfo uriInfo;

    public AsyncInvocationInterceptor newInterceptor() {
        return this;
    }

    public void prepareContext() {
        LOG.finest("invoking request for {0} on original thread {1}", uriInfo.getPath(), Thread.currentThread().getId());
    }

    public void applyContext() {
        LOG.finest("receiving response for {0} on async thread {1}", uriInfo.getPath(), Thread.currentThread().getId());
    }
}
```

Is there a simpler way to enable both of these use cases?

spec/src/main/asciidoc/clientexamples.asciidoc
```diff
@@ -146,3 +146,13 @@ public interface BadInterfaceThree {
...
 `BadInterfaceThree` has a mismatch.  The `@Path` annotation declares a URI template named "someOtherParam" but the `@PathParam` annotation specifies a template named "notTheSameParam".
 All three interfaces will result in a RestClientDefinitionException.
+
+Implementations must throw a RestClientDefinitionException if an asynchronous method that accepts an InvocationCallback parameter also has a non-void return type.  For example:
+[source, java]
+----
+@Path("/somePath")
+public interface BadAsyncInterface {
+    @GET
+    public Response getEntry(InvocationCallback<SomeEntry> callback);
```
kenfinnigan: This should be updated as we don't support `InvocationCallback` anymore

andymc12: Nice catch!  I'm fixing that now.

andymc12:
I'm planning to merge this PR now, but I will leave Issue #10 open a little longer in case we want to make further refinements/improvements to async method support.

spec/src/main/asciidoc/async.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+//
+
+[[restasync]]
+== MicroProfile Rest Client Asynchronous Support
+
+It is possible for Rest Client interface methods to be declared asynchronous.  This allows the thread invoking the interface method to proceed while the RESTful request occurs on another thread.
+
+=== Asynchronous Methods
+
+A method is considered to be asynchronous if either of the method's return type is `java.util.concurrent.CompletableFuture`.
```
Ladicek: "either of the method's return type" sounds wrong, should be just "the method's return type".

spec/src/main/asciidoc/async.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+
+There may be cases where it is necessary for client application code or runtime components to be notified when control of the client request/response has flowed from one thread to another.
+This can be accomplished by registering an implementation of the AsyncInvocationInterceptorFactory provider interface.
+MP Rest Client implementations must invoke the `newInterceptor` method of each registered factory provider prior to switching thread of execution on async method requests.
+That method will return an instance of `AsyncInvocationInterceptor` - the MP Rest Client implementation must then invoke the `prepareContext` method while still executing on the thread that invoked the async method.
+After swapping threads, but before invoking further providers or returning control back to the async method caller, the MP Rest Client implementation must invoke the `applyContext` method on the new async thread that will complete the request/response.
+
+The following example shows how the AsyncInvocationInterceptorFactory provider and associated AsyncInvocationInterceptor interface could be used to propagate a `ThreadLocal` value from the originating thread to async thread:
```
Ladicek: The `someValue` field isn't really `ThreadLocal`; I guess the intent here is "ThreadLocal-like" or something like that?

spec/src/main/asciidoc/async.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+----
+public class MyFactory implements AsyncInvocationInterceptorFactory {
+
+    public AsyncInvocationInterceptor newInterceptor() {
+         return new MyInterceptor();
+    }
+}
+
+public class MyInterceptor implements AsyncInvocationInterceptor {
+    String someValue;
```
Ladicek: Is there a visibility guarantee mandated by the spec, or should this field be `volatile`?

Ladicek:
I added a few relatively minor comments; sorry for not getting to it earlier!

spec/src/main/asciidoc/async.asciidoc
andymc12: Right.  The `someValue` field is only used to copy the value from one thread to the other.  Would it help make the example more clear if I added the code for `SomeClass` (which contains the actual `ThreadLocal`?

spec/src/main/asciidoc/async.asciidoc
andymc12: Good catch - it should be `volatile` - I will change that.

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -101,6 +102,19 @@ default RestClientBuilder baseUri(URI uri) {
...
 
+    /**
+     * Specifies the <code>ExecutorService</code> to use when invoking
+     * asynchronous Rest Client interface methods.  By default, the executor
+     * service used is determined by the MP Rest Client implementation runtime.
+     *
+     * @param executor the executor service for the runtime to use when invoking
+     * asynchronous Rest Client interface methods - must be non-null.
+     * @return the current builder with the executorService set.
+     * @throws NullPointerException if the <code>executor</code> parameter is
```
Emily-Jiang: I would suggest to use IllegalArgumentException instead of NPE.

spec/src/main/asciidoc/async.asciidoc
Ladicek: Ah I see! I didn't notice the content of the `prepare` and `apply` methods :-) I think this is fine, sorry for the noise.

Issue #86, opened 2018-03-08
Need package-level javadoc description


andymc12:
The package level javadoc description for the 1.0.X and dev streams have no description in the Javadocs - http://download.eclipse.org/microprofile/microprofile-rest-client-1.0.1/apidocs/

specifically:
http://download.eclipse.org/microprofile/microprofile-rest-client-1.0.1/apidocs/org/eclipse/microprofile/rest/client/package-summary.html (only lists Emily as the author)
all the rest have no description at all, for example:
http://download.eclipse.org/microprofile/microprofile-rest-client-1.0.1/apidocs/org/eclipse/microprofile/rest/client/annotation/package-summary.html

This should be changed by adding javadoc comments to the package-info.java files.

andymc12:
Resolved by #87

PR #87, opened 2018-03-08
Add javadoc to package-info.java

andymc12:
This adds javadoc text with examples to package-info.java files, allowing users and implementors to see more information at the package level.

I am targeting this change toward the 1.0.X-service branch, but it should also go into master.

This will resolve Issue #86.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/package-info.java
@@ -19,7 +19,24 @@
  *******************************************************************************/
 
 /**
+ * APIs for extending MP Rest Client functionality - such as new providers.
+ * Example:
+ * <pre>
+ * public interface MyClientService {
+ *     @GET
+ *     @Path("/myService/{id}")
+ *     Widget getWidget(@PathParam("id") String id) throws UnknownWidgetException;
+ * }
+ * ...
+ * public class UnknownWidgetExceptionMapper implements ResponseExceptionMapper {
  *
+ *     @Override
+ *     UnknownWidgetException toThrowable(Response response) {
+ *         String msg = "Could not find widget with ID of " + response.getHeaderString("WidgetId");
+ *         return new UnknownWidgetException(msg)
+ *     }
+ * }
+ * </pre>
  */
 @org.osgi.annotation.versioning.Version("1.0.1")
 package org.eclipse.microprofile.rest.client.ext;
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/ext/package-info.java
```diff
@@ -19,7 +19,24 @@
...
+ * APIs for extending MP Rest Client functionality - such as new providers.
+ * Example:
+ * <pre>
+ * public interface MyClientService {
+ *     @GET
+ *     @Path("/myService/{id}")
+ *     Widget getWidget(@PathParam("id") String id) throws UnknownWidgetException;
+ * }
+ * ...
+ * public class UnknownWidgetExceptionMapper implements ResponseExceptionMapper {
```
aguibert: Is it correct that a custom `ResponseExceptionMapper` doesn't require a `@RegisterProvider` annotation?  Seems odd that this can be figured out automatically but MBR, MBW, and RequestFilter require the anno.

api/src/main/java/org/eclipse/microprofile/rest/client/inject/package-info.java
```diff
@@ -19,7 +19,30 @@
...
 /**
+ * APIs to aid in CDI-based injection of MP Rest Client implementations.  These
+ * annotations are used both to mark an interface as registered Rest Client and
+ * also to designate that an implementation of that interface should be injected
+ * at a specific injection point.
  *
+ * Example:
+ * <pre>
+ * @RegisterProvider
+ * @Dependent
```
aguibert: Why is `@Dependent` needed?  Can't `@RegisterProvider` be a bean-defining annotation?

api/src/main/java/org/eclipse/microprofile/rest/client/ext/package-info.java
andymc12: The `@RegisterProvider` annotation is not always required on the client interface - it would require that the mapper be registered using the `RestClientBuilder` API.  But since that is not obvious here, I'll update the interface to add the annotation.  Thanks!

api/src/main/java/org/eclipse/microprofile/rest/client/inject/package-info.java
andymc12: `@Dependent` is a bean-defining annotation for MP Rest Client 1.1, but this javadoc is intended for both 1.1 and 1.0 (where it is not required to be a bean-defining annotation).  We could remove `@Dependent` from the 1.1 javadocs, but that should probably be a part of PR #83.

PR #91, opened 2018-04-16
Issue #10: Use CompletionStage rather than CompletableFuture

andymc12:
Updates to spec and TCK to use the CompletionStage interface
rather than the concrete CompletableFuture class.  This change
was prompted by a discussion in the gitter.im chat.  It more
closely aligns to the JAX-RS 2.1 reactive client APIs as well
as changes proposed in the MP Fault Tolerance spec.

### DIFF
--- a/spec/src/main/asciidoc/async.asciidoc
+++ b/spec/src/main/asciidoc/async.asciidoc
@@ -21,7 +21,7 @@ It is possible for Rest Client interface methods to be declared asynchronous.  T
 
 === Asynchronous Methods
 
-A method is considered to be asynchronous if the method's return type is `java.util.concurrent.CompletableFuture`.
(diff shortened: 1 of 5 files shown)


spec/src/main/asciidoc/async.asciidoc
```diff
@@ -21,7 +21,7 @@ It is possible for Rest Client interface methods to be declared asynchronous.  T
 
 === Asynchronous Methods
 
-A method is considered to be asynchronous if the method's return type is `java.util.concurrent.CompletableFuture`.
+A method is considered to be asynchronous if the method's return type is `java.util.concurrent.CompletionStage`.
```
OndrejM: Just a thought - could we support all return types that implement CompletionStage, including CompletableFuture? It's not necessary and we can discuss later, I just wanted to know if anybody knows about some drawbacks to doing this...

PR #92, opened 2018-04-19
[TCK] Entity must not be null for http method PUT

jGauravGupta:


andymc12:
Hi @jGauravGupta I don't think that there is a requirement that entities must be non-null for PUT requests, but I think this change is fine - adding a body to the request (and then verifying it in the response) is good.

andymc12:
Hi @OndrejM - yes, I can merge this.  I wanted to make sure that others had plenty of time to register any objections/approvals, but since it has been longer than a week I think we can merge it now.

Thanks Ondrej, and thanks @jGauravGupta for the patch!

Issue #74, opened 2018-02-06
Arguable use of ParamConverterProvider


heiko-braun:
Some tests (see discussion on https://github.com/eclipse/microprofile-rest-client/pull/69) make use of ParamConverterProvider to change the request parameter values on the client side, prior to the actual HTTP request.

I think using a ParamConverter for these purposes is arguable or even wrong.  The spec only references ParamConverter for injection cases to supply values to resource fields annotated with `@<X>Param`. See section 3.2 in https://github.com/jax-rs/spec/blob/master/spec.pdf.

IMO the following tests make invalid use of a `ParamConverterProvider` and should be changed to not depend on the a `ParamConverter`, which seems to be a server-side extension point only:

- `InvokeWithBuiltProvidersTest` 
-  `InvokeWithRegisteredProvidersTest`
-  `CDIInvokeWithRegisteredProvidersTest` 

@andymc12 Thoughts?

andymc12:
@heiko-braun I agree that it is a little strange expecting a ParamConverter to convert a String to a String.  A better example might be to convert some custom object to String and vice-versa, which I believe is the intent of the ParamConverters.  

I still think it makes sense for ParamConverters to exist on the client side as it facilitates type-safety and better usage of model objects - for example:

```
public interface InventoryService {
    @POST
    @Path("/{widgetId}/{quantity}")
    void order(@PathParam("widgetId") Widget widget, @PathParam("quantity") int qty);

    // vs

    @POST
    @Path("/{widgetId}/{quantity}")
    void order(@PathParam("widgetId") String widgetId, @PathParam("quantity") int qty);
}

... // and

inventory.order(widget, 100);
// vs
inventory.order(widget.getId(), 100); // or some other method to get the id of the widget in question

```

With ParamConverters, users can write the business logic with less code.  What would you think if we changed the ParamConverters and the tests that use them to use custom objects that convert to and from Strings?

heiko-braun:
Thanks @andymc12, I can see the use case. It makes sense when you don't control the Java interface to begin with. I guess I was looking from the angle of creating the interface myself and wondering about the additional complexity. 

The other thing worth mentioning, is the asymmetry. The  `ParamConverter` makes it appear as if the conversion would be happening in both ways, inbound and outbound. But in fact it's really just outbound. On the server-side it's the other way around, but also just in one direction.  But that might be just my RPC wired brain, that took a while to understand what's going on while implementing it.

[...]

PR #93, opened 2018-04-19
[TCK] PUT request input path and stub url mismatch cause HTTP 404 Not Found

jGauravGupta:


kenfinnigan:
Can you elaborate on the issue, as I believe both Apache CXF and WF Swarm are passing the TCK with the test as it stands

andymc12:
I think the issue here is the same as discussed in issue #74 and #69 - it sounds like your implementation is not converting the parameters correctly.  This test case should be valid - though we have considered changing the ParamConverter so that it actually converts a custom object to a string, rather than converting string to string - but this should be considered an enhancement to the TCK (mainly to make the test more real-world-like).

Assuming that this is the issue you are seeing, I would recommend investigating your implementation to make sure that the `TestParamConverter` is getting invoked correctly.  Then this test case should pass for you.  

That said, if you wanted to convert the test so that it would be more real-world-like - i.e. converting some type of custom object to a string in the ParamConverter, that would be very much appreciated!

Thanks, Andy

jGauravGupta:
Hi @andymc12 ,

Thanks for highlighting the issue #74 and #69, yes it's a issue with Jersey client.

PR #94, opened 2018-04-20
TCK improvement

jGauravGupta:


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/CDIURIvsURLConfigTest.java
```diff
@@ -43,17 +43,18 @@
...
+
     @Inject
     @RestClient
     private SimpleGetApi api;
 
     @Deployment
     public static WebArchive createDeployment() {
         String uriPropertyName = SimpleGetApi.class.getName()+"/mp-rest/uri";
-        String uriValue = getStringURL() + "/uri";
+        String uriValue = getStringURL() + "uri";
```
jGauravGupta: getStringURL() returns url with slash at the end "http://localhost:8765/"

OndrejM: getStringURL() doesn't ensure that the URL is always returned with `/` at the end. I would also fix the getStringURL() method to append `/` if it's missing

PR #108, opened 2018-07-09
Resolve issue where 1.1 TCK tests invoke the same URI twice

andymc12:
Two TCK methods both check that the "/" URI was executed exactly once - since wiremock does not reset in between test methods, the second test method to execute will always fail. This change uses a different URI for the executorService test to ensure that there is no URI conflicts between tests.

This pull request is agains the 1.1.X-service branch, but the same change should also be made in master.


andymc12:
It is possible to workaround this issue by providing a testng.xml file to run the two test cases separately, like: 
https://github.com/andymc12/cxf/blob/ec68ee7cde32c2b449eb742f4d7e2f5a8323f5da/systests/microprofile/client/weld/testng.xml

Issue #121, opened 2018-09-14
Make interface type available to providers


pavolloffay:
Use case:
Provider could use annotations on the interface to configure its behavior. For example tracing integration could be disabled if `@Traced(false)` is added on the interface. Or it could add custom meta-data - tags specified in that annotation to the span created in the client.

This is blocking https://github.com/eclipse/microprofile-opentracing/pull/102

kenfinnigan:
Where would such information be made available?

Not sure I'm understanding what's required and how.

pavolloffay:
Let's say I have this client:

```java
@Traced
@TracedTags({"foo": "bar"})
@Path("/")
interface Service {
  @GET
  @Path("/get")
   get();
}
```

It would mean that the client is traced and the tracing integration would add `foo:bar` to the span created for the method invocation. Generally speaking there could be any annotation, important is that the `Service.class` should be available to the provider/feature class which is added to the client.

The `Service.class` interface could be made available in 
`javax.ws.rs.core.Coonfiguration.getProperty("mp.rest.client.interface")`. Which is accessible in `Feature` and `ClientRequestContext`.

andymc12:
Talking with @pavolloffay - it sounds like the ideal solution would be to have a `RestClientListener` that is invoked with the client itself is created (as opposed to the `RestClientBuilderListener` which is invoked when the builder is created).  The `RestClientListener`'s `onNewClient` method should include the rest client interface class - something like:
 `void onNewClient(Class<?> clientInterface, RestClientBuilder builder)`

pavolloffay:
@andymc12 I have more thoughts on the issue. 

It would be great if a provider(e.g. filter) would have access to the method of the interface(which means it has an access to the interface itself).  Then an annotation can be used to change a behavior of the method e.g. `@Traced(false)` on the specific method rather than on the whole client.

andymc12:
@pavolloffay @kenfinnigan I think this request makes sense.  What do you guys think of something like:

```
public interface RestClientRequestContext extends ClientRequestContext {

    Method getInvokedMethod();
}

...

public interface RestClientRequestFilter {

    void filter(RestClientRequestContext requestContext) throws IOException;
}
```

pavolloffay:
[...]

PR #124, opened 2018-09-20
Issue 121: RestClientBuilder SPI

andymc12:
This enables global registration of providers, etc. during the creation of rest client instances (as opposed to builders). It also passes in the rest client service interface allowing implementations to perform client-specific computations.

This resolves issue #121.


pavolloffay:
@andymc12 see my last comment in https://github.com/eclipse/microprofile-rest-client/issues/121#issuecomment-424009792.

I think it would be better to pass the invoking method of the client interface into the request properties. This way a provider has full access to annotations on the method and class. Base on this annotation it can change behaviour.

Issue #90, opened 2018-04-07
Add url and other attributes to the RestClient qualifier


OndrejM:
Current version 1.0.1 requires that the URL and other attributes like `scope` and `providers` for an injected client proxy is specified using an external MP Config value under a generated config key.

I suggest to allow specifying the URL also in the code as the default value, using a non-binding attribute of the RestClient qualifier, such as:

```
@Inject
@RestClient(url="http://localhost/myservice", providers = {MyProvider.class, AnotherProvider.class})
MyService myService
```

kenfinnigan:
@OndrejM, any reason for preferring the url be specified on the injection point vs the actual interface?

My concern is the same service being used in different places requires the url to be duplicated in string form

andymc12:
@kenfinnigan and I talked about this on the call.  Is there a reason you'd like to specify the URL in the injection point?  One alternative we discussed would be to specify it on the interface - like:
```
@RegisterRestClient(url="http://localhost/myservice")
public interface MyService {...
```

It is already possible to specify the providers using `@RegisterProvider` annotations on the interface.

Thanks, Andy

kenfinnigan:
@OndrejM, is the main problem you're raising here related to rapid development and iterations? ie. needing to define MP Config at all.

Or is it more around knowing what the generated config key needs to be?

OndroMih:
I agree that the baseUri attribute on @RegisterRestClient is better. And even better that since version 1.3 it's possible to define a custom config key (#115) to supply the URL using MP Config.

PR #131, opened 2018-11-05
Issue 90: Allow URI to be set in `@RegisterRestClient` annotation

andymc12:
This issue resolves issue #90 though there may be other attributes we could add to the `@RegisterRestClient` annotations (one suggestion was providers but they can already be declared on the interface via `@RegisterProvider` annotations.  


andymc12:
Thanks for the review Ken!

PR #137, opened 2018-12-03
Added a TCK test for reusing a client after failure, fixes #128

### LINKED ISSUE #128: TCK to reuse the client
I could not find TCK which reuses the same client for multiple requests, especially for failed requests 5xx. Such a test should be added to TCK.

michalszynkiewicz:


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientReuseTest.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
andymc12: Can you change this to 2018?

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientReuseTest.java
```diff
@@ -0,0 +1,121 @@
...
+        performSuccessfulRequest(client);
+
+        performFailingRequest(client);
+
+        performSuccessfulRequest(client);
+        return null;
+    }
+
+    private void performSuccessfulRequest(JsonPClient client) {
+        stubReturning("{\"content\": true}");
```
andymc12: For both of the calls to `stubReturning` (here and in the `performFailingRequest` method), I ran into problems in my local environment.  I think WireMock must use thread locals that are not copied over to the thread in the ExecutorService.  I was able to make this work by calling `stubReturning` from the main test method (`shouldReuseClientAfterFailure`) before calling the `callWithTimeout` method.  This also required that I use different URIs (IDs) so that I could set them all in advance.  Here is a short summary of my changes:

```
    @Test
    public void shouldReuseClientAfterFailure() throws Throwable {
        stubReturning(1, "{\"content\": true}");
        stubReturning(2, "Not a json");
        stubReturning(3, "{\"content2\": true}");
        callWithTimeout(this::testReuseClientAfterFailure, 20, TimeUnit.SECONDS);
    }

    private Void testReuseClientAfterFailure() {
        JsonPClient client =
            RestClientBuilder.newBuilder()
                .baseUri(getServerURI())
                .build(JsonPClient.class);

        //expect successful invocation
        assertTrue(client.get("1").getBoolean("content"));

        //expect failing invocation
        expectFailure(() -> client.get("2"));

        //expect successful invocation
        assertTrue(client.get("3").getBoolean("content2"));
        return null;
    }

    private void stubReturning(int id, String text) {
        stubFor(
            get(urlEqualTo("/" + id))
                .willReturn(
                    aResponse().withBody(text).withHeader("Content-Type", "application/json")
                )
        );
    }
...
```

With these changes, this test passes for me in CXF.

michalszynkiewicz: @andymc12 thanks. It works as it is for SmallRye & RestEasy but we're using slightly newer versions of testng, etc, so this may be to blame. 
I will update it.

andymc12: That could be, or possibly SmallRye/RestEasy are testing with wiremock on the default port (8080).  I noticed in the CXF testing environment, that the wiremock client was attempting to connect on port 8080 (I'm guess that that is the default port), but pom file specified port 8765, so I was getting a connect exception.

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientReuseTest.java
michalszynkiewicz: done

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientReuseTest.java
michalszynkiewicz: done

andymc12:
Looks good, thanks @michalszynkiewicz !!

andymc12:
This should resolve issue #128

PR #141, opened 2018-12-10
modified ProducesConsumesTest to pass payload

michalszynkiewicz:
The current version of `ProducesConsumesTest` checks if `@Consumes` annotation on a method that doesn't send any payload results in passing an expected content type header.
Passing content type for a request with no payload doesn't seem to make much sense to me, and this is also the assumption of Rest Easy which makes the test in the current form fail for SmallRye.

This change modifies the test to pass entities for checks for `@Consumes`


michalszynkiewicz:
I added interface-level annotations and a test that checks if they are used if content types are not specified on the method level. Is this what you meant?
And it works with RestEasy :)

Issue #116, opened 2018-08-29
Proposal for setting and propagating HTTP Header values


kenfinnigan:
@andymc12 and I discussed this problem at length in today's meeting, which is also referenced as part of #73 and #104. I've created a separate issue to be able to outline some thoughts we came up with today in an effort to garner some additional feedback on the approach.

The spec would define an interface for setting/propagating headers such as:
```
public interface ClientHeaders {
  MultivaluedMap set(MultivaluedMap incomingHeaders, MultivaluedMap clientOutgoingHeaders);
}
```
The purpose of which is to pass in the map of headers from an incoming request and the map of headers that will be set on the outgoing client request based on any annotations that are defined on the service interface, such as `@HeaderParam`. The second parameter, `clientOutgoingHeaders` would be read only in this method. The map returned from this method would then be merged with the map that was passed into `clientOutgoingHeaders` to create the full set of headers for the client request. We pass in the currently defined outgoing headers to provide an opportunity to adjust the values if necessary.

In addition to the interface above, noting that `ClientHeaders` is just a name I came up with not what it will necessarily end up being called, the spec would provide a concrete implementation of the interface that offers the ability to propagate headers from incoming to outgoing based on values within a specific config property. In the short term we may consider adding the headers into a single MP REST Client config property, but longer term it would be better for MP specs to define their own list of headers to propagate that could be read by this concrete class.

Developers then have the ability to define their own classes to perform setting and propagation of headers, but also could extend the spec implementation to include additional header setting that may not be defined by the spec.

So how would this be used?

The spec would define an annotation, possibly called `@ClientHeaderParams`, that can be used on the injection point with `@RESTClient`. By default this annotation would use the concrete implementation from the spec to propagate headers to the client request. If the developer has their own implementation of `ClientHeaders`, they can specify the class on the annotation and that class will be used to propagate/set headers.

In addition, we'd likely add something like `@ClientHeaderParam`, which would take a key and value for setting specific values on the new client request. It would work in a similar fashion to `@HeaderParam`.

We felt it was necessary to offer a way through a method call, `ClientHeaders`, to set a group of header properties as adding a `@ClientHeaderParam` for everything to be set/propagated could be quite lengthy, especially when dealing with tracing headers.

I think I've covered the main pieces we discussed, please correct me @andymc12 if I missed anything.

Please provide feedback on the idea.

chunlongliang-ibm:
[...]

PR #143, opened 2018-12-13
[116] New `@ClientHeaderParam` annotation and TCK tests

andymc12:
Includes spec document changes describing how to use `@ClientHeaderParam`, the annotation in the API (including javadoc) and TCK tests that test proper usage and exception conditions.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.rest.client.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to specify an HTTP parameter that should be sent with the outbound request.
+ * When this annotation is placed at the interface level of a REST client interface, the specified header will be sent on each request for all
+ * methods in the interface.
+ * When this annotation is placed on a method, the header will be sent only for that method. If the same HTTP header is specified in an annotation
+ * for both the type and the method, only the header value specified in the annotation on the method will be sent.
+ * <p>
+ * The value of the header to send can be specified explicitly by using the <code>value</code> attribute.
+ * The value can also be computed via a default method on the client interface that returns a String value.  This method must be specified in the
+ * <code>value</code> attribute but wrapped in curly-braces.
+ * The default method's signature must either contain no arguments or a single <code>String</code> argument. The String argument is the name of the
+ * header.
+ * <p>
+ * Here is an example using both approaches:
+ * <pre>
+ * public interface MyClient {
+ *
+ *    static AtomicInteger counter = new AtomicInteger(1);
+ *
+ *    default String determineHeaderValue(String headerName) {
+ *        if ("SomeHeader".equals(headerName)) {
+ *            return "InvokedCount " + counter.getAndIncrement();
+ *        }
+ *        throw new UnsupportedOperationException("unknown header name");
+ *    }
+ *
+ *    {@literal @}ClientHeaderParam(name="SomeHeader", value="ExplicitlyDefinedValue")
+ *    {@literal @}GET
+ *    Response useExplicitHeaderValue();
+ *
+ *    {@literal @}ClientHeaderParam(name="SomeHeader", value="{determineHeaderValue}")
+ *    {@literal @}DELETE
+ *    Response useComputedHeaderValue();
+ * }
+ * </pre>
+ * The implementation should fail to deploy a client interface if the annotation contains a <code>@ClientHeaderParam</code> annotation with a
+ * <code>value</code> attribute that references a method that does not exist, or contains an invalid signature.
+ * <p>
+ * The <code>required</code> attribute will determine what action the implementation should take if the method specified in the <code>value</code>
+ * attribute throws an exception. If the attribute is true (default), then the implementation will abort the request and will throw the exception
+ * back to the caller. If the <code>required</code> attribute is set to false, then the implementation will not send this header if the method throws
+ * an exception.
+ * <p>
+ * Note that if an interface method contains an argument annotated with <code>@HeaderParam</code>, that argument will take priority over anything
+ * specified in a ClientHeaderParam annotation.
+ *
+ * @since 1.2
+ */
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Repeatable(ClientHeaderParams.class)
+public @interface ClientHeaderParam {
+    /**
+     * @return the name of the HTTP header.
+     */
+    String name();
+
+    /**
+     * @return the value of the HTTP header - or the method to invoke to get the value (surrounded by curly braces).
+     */
+    String value();
+
+    /**
+     * @return whether to abort the request if the method to compute the header value throws an exception (true; default) or just skip this header
+     * (false)
+     */
+    boolean required() default true;
+}
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
@@ -31,12 +31,12 @@ import java.lang.annotation.Target;
  * for both the type and the method, only the header value specified in the annotation on the method will be sent.
  * <p>
  * The value of the header to send can be specified explicitly by using the <code>value</code> attribute.
- * The value can also be computed via a default method on the client interface that returns a String value.  This method must be specified in the
- * <code>value</code> attribute but wrapped in curly-braces.
- * The default method's signature must either contain no arguments or a single <code>String</code> argument. The String argument is the name of the
- * header.
+ * The value can also be computed via a default method on the client interface or a public static method on a different class.  The compute method
+ * must return a String or String[] (indicating a multivalued header) value.  This method must be specified in the <code>value</code> attribute but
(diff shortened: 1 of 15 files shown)


andymc12:
This is part 1 of the solution for issue #116.

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
```diff
@@ -0,0 +1,94 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to specify an HTTP parameter that should be sent with the outbound request.
+ * When this annotation is placed at the interface level of a REST client interface, the specified header will be sent on each request for all
+ * methods in the interface.
+ * When this annotation is placed on a method, the header will be sent only for that method. If the same HTTP header is specified in an annotation
+ * for both the type and the method, only the header value specified in the annotation on the method will be sent.
```
kenfinnigan: Does this need to cater to header params that are lists?

For instance, if the annotation is present on interface and method, if the type is a list then we add them both?

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
```diff
@@ -0,0 +1,94 @@
...
+ * <p>
+ * The value of the header to send can be specified explicitly by using the <code>value</code> attribute.
+ * The value can also be computed via a default method on the client interface that returns a String value.  This method must be specified in the
```
kenfinnigan: Are we also supporting {MyClass.myMethod} usage, or only default methods?

spec/src/main/asciidoc/clientexamples.asciidoc
```diff
@@ -80,6 +80,49 @@ All built in HTTP methods are supported by the client API.  Likewise, all base p
...
+The `@ClientHeaderParam` annotation can allow users to specify HTTP headers that should be sent without altering the client interface method signature. The annotation contains three attributes: `name`, `value`. and `required`.  The name attribute is used to specify the header name. The value attribute is used to specify the value of the header. The value can be specified explicitly
+or it can reference a default method on the client interface that would compute the value of the header - in this latter case, the compute method name must be surrounded by curly braces. The compute method must be a default method on the interface, must return a `String` and must either contain no arguments or contain a single `String` argument - the implementation will use this
+`String` argument to pass the name of the header. The required attribute determines what should happen in the event that the compute method throws an exception. If the required attribute is set to true (default), then the client request will fail if the compute method throws an exception.  If it is set to false, then the client request will continue, but without sending the HTTP header.
```
kenfinnigan: Do we need some kind of big flashing warning in the spec doc about the fact that required=true will cause an exception if the method call fails?

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
andymc12: I think that an annotation on the method should always override an annotation on the interface - this is consistent with other patterns (like Java EE's security annotations).

But, I think that you make a good point about headers that are lists.  We probably should make the value of the annotation a `String[]` rather than a `String`.

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
andymc12: My thought was only default methods.  That still gives users flexibility with extending other interfaces, etc.  Do you think we need to support {MyClass.myMethod} semantics?

kenfinnigan: I think in the call we discussed utility methods too, so that you could have a common method to retrieve an authentication token, for instance, without having to replicate that code across all REST Client interfaces.

You could have the default method in multiple interfaces call that utility method, but that just seems like unnecessary boilerplate

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
kenfinnigan: Could we not just recognize it's a multi valued header and combine them?

spec/src/main/asciidoc/clientexamples.asciidoc
andymc12: Sure, I can bold that text.

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
andymc12: I think it would add complexity.  Potentially any header could have multiple values, so then we'd run into issues where it isn't clear whether a user is trying to override a header value in the method or whether they are trying to append another value.  We could add some sort of "appendable" attribute to the annotation to indicate that the value should be appended if multiple headers are already defined, but I think that makes it more complicated - and potentially unwieldy in a case like this:

```
@ClientHeaderParam(name="MyHeader", value="FromType")
public interface MyClient {

    @ClientHeaderParam(name="MyHeader", value="FromMethod")
    Response myMethod(@HeaderParam("MyHeader") passedInHeader);
}
...
myClient.myMethod("FromMethodArgument");
```

This could lead to sending:
`MyHeader: FromMethodArgument,FromType,FromMethod`
`MyHeader: FromType,FromMethod` (only using the ClientHeaderParams, ignoring the HeaderParam)
`MyHeader: FromMethodArgument` (only using the HeaderParam, ignoring the ClientHeaderParams)
or perhaps other combinations.

I was thinking that to make this as simple as possible, `@HeaderParam` takes priority over the `@ClientHeaderParam`, since that is passed-in by the user - part of the app.  Then the `@ClientHeaderParam` annotation on method would take priority over the annotation on the interface, since it is more specific.

So if a user wants multiple values for a particular header, then I think the annotation should support it with a `String[]` value attribute (and allow the default method to return a String array).  Alternatively, we could use multiple annotations at the same level - something like this:
```
@ClientHeaderParam(name="header1", value="value1")
@ClientHeaderParam(name="header1", value="value2")
@ClientHeaderParam(name="header1", value="value3")
Response myMethod();
```

but I prefer a String array attribute.

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
andymc12: That makes sense.  So then the requirement for the compute method would be that it must either be a default method on the client interface, or a static method that is visible to the interface (i.e. not private, or package-protected in another package, etc.).

kenfinnigan: Sounds good

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
kenfinnigan: Ok, I'd forgot about the override case.

So I'm good with making it a `String[]` and then following the original override means

andymc12:
@kenfinnigan @michalszynkiewicz - can you take another look at this PR?  I've updated it with the review comments, including:
    - Highlighting `required` attribute behavior
    - Making `value` attribute a String[]
    - Enabling compute method to be from a different class
    - Validation failure if same header specified twice on same target

Thanks!

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/InvalidComputeMethodSignature.java
```diff
@@ -0,0 +1,52 @@
...
+
+    default String invalidMethod(String headerName, String someOtherArg) {
+        return "should not be invoked - too many String args";
+    }
+
+    default String invalidMethod(String headerName, Integer someOtherArg) {
+        return "should not be invoked - unexpected Integer arg";
+    }
+
+    default String invalidMethod(ClientRequestContext context, ClientRequestContext someOtherArg) {
```
michalszynkiewicz: should it work with single `ClientRequestContext`?
I cannot find a test that would check it.

spec/src/main/asciidoc/clientexamples.asciidoc
```diff
@@ -80,6 +80,61 @@ All built in HTTP methods are supported by the client API.  Likewise, all base p
...
+The `@ClientHeaderParam` annotation can allow users to specify HTTP headers that should be sent without altering the client interface method signature. The annotation contains three attributes: `name`, `value`. and `required`.  The name attribute is used to specify the header name. The value attribute is used to specify the value(s) of the header. The value can be specified explicitly
+or it can reference a method that would compute the value of the header - in this latter case, the compute method name must be surrounded by curly braces. The compute method must be either a default method on the interface or a public static method that is accessible to the interface, must return a `String` or `String[]` and must either contain no arguments or contain a single `String`
+argument - the implementation will use this `String` argument to pass the name of the header. When specifying a compute method as the value attribute, only one method may be specified - if more than one string is specified as the value attribute, and one of the strings is a compute method (surrounded by curly braces), then the implementation will throw a RestClientDefinitionException.
```
michalszynkiewicz: It would be good to also have a TCK test checking if `RCDException` is thrown with the case of "more than one string is specified as the value attribute, and one of the strings is a compute method"

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/InvalidComputeMethodSignature.java
andymc12: I was originally thinking that we could pass in a `ClientRequestContext` in addition the header name, but ultimately decided against it since a lot of the info in CRC would be invalid (or at least, not yet determined this early in the flow).  

I should remove that method from the TCK interface.

Is there anything in particular that you thought might be useful from the CRC?  We could possibly pass in the `Method` object from the client interface - that might allow compute methods to read other annotations, etc to better determine what the header value should be...

spec/src/main/asciidoc/clientexamples.asciidoc
andymc12: Agreed.  I'll add this check.

api/src/main/java/org/eclipse/microprofile/rest/client/annotation/ClientHeaderParam.java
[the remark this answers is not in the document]
andymc12: Good catch.  I'll update this to be more clear that:
1) The method can be either a default method on the rest client interface or a public static utility method.
and
2) That the compute method can return String or String[].

andymc12:
@michalszynkiewicz I updated the Javadoc to be consistent with the spec text to indicate that the compute method can be either a default method on the interface _or_ a public static method on another class/interface.  And I added a new test for cases where multiple values were specified on the same annotation and at least one uses the curly braces.  Thanks again for the review!

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/InvalidComputeMethodSignature.java
michalszynkiewicz: I didn't have anything specific in mind.

andymc12:
@kenfinnigan I assume you are ok with the latest changes (based on discussion in gitter from a couple weeks ago).  I'll plan to merge tomorrow unless you have any objections.  Thanks (and Happy New Year)!

PR #145, opened 2018-12-19
[142] test timeouts with CDI

### LINKED ISSUE #142: TCK expects MP Config settings applied to non-CDI defined interfaces
`TimeoutViaMPConfigTest` expects `connectTimeout` and `readTimeout` from MP Config settings to set the appropriate timeouts on a client that is created with `RestClientBuilder`, and, if I understand correctly, to take precedence over the builder setting.

While it may be reasonable to let users override what they set programmatically with configuration properties, the spec says "For CDI defined interfaces, it is possible to use MicroProfile Config properties to define...", and doesn't seem to mention that these properties should also work with a programmatically configured client.

@andymc12 am I missing something? If it's a bug in the TCK, I can try to create a PR to fix it.

https://github.com/eclipse/microprofile-rest-client/blob/450f6d5c01fe8f3b21eaf223be713b118b4f5427/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/TimeoutTestBase.java#L52-L67

and:
https://github.com/eclipse/microprofile-rest-client/blob/450f6d5c01fe8f3b21eaf223be713b118b4f5427/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/TimeoutViaMPConfigTest.java#L34-L52

michalszynkiewicz:
+ test that MP Config timeouts don't impact programmatically created clients
+ rename of the async CDIInvokeSimpleGetOperationTest to fix intermittent failures caused by the name clash

fixes #142

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/timeout/TimeoutBuilderIndependentOfMPConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/timeout/TimeoutBuilderIndependentOfMPConfigTest.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2018 Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.rest.client.tck.timeout;
+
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=142

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=142

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/timeout/TimeoutBuilderIndependentOfMPConfigTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.testng.Assert.assertTrue;
+
+public class TimeoutBuilderIndependentOfMPConfigTest extends TimeoutTestBase {
+
+    @Deployment
+    public static Archive<?> createDeployment() {
+        String clientName = SimpleGetApi.class.getName();
+        String timeoutProps = clientName + "/mp-rest/connectTimeout=7000" +
```
michalszynkiewicz: change to 15s in the config settings

PR #159, opened 2019-02-07
modified CliHeadParamTest to cover http headers sent in multiple lines

michalszynkiewicz:


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientHeaderParamTest.java
```diff
@@ -28,8 +28,9 @@
 import static org.testng.Assert.fail;
 
 import java.util.Arrays;
-import java.util.Map;
 import java.util.stream.Collectors;
```
andymc12: We still need to import `Map` since it is used in `testHeaderNotSentWhenExceptionThrownAndRequiredIsFalse`

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientHeaderParamTest.java
```diff
@@ -66,9 +67,14 @@ private static void stub(String expectedHeaderName, String... expectedHeaderValu
         String expectedIncomingHeader = Arrays.stream(expectedHeaderValue)
                                               .collect(Collectors.joining(","));
         String outputBody = expectedIncomingHeader.replace(',', '-');
+        MappingBuilder mappingBuilder = get(urlEqualTo("/"));
+
+        // headers can be sent either in a single line with comma-separated values o in multiple lines
```
andymc12:
```suggestion
        // headers can be sent either in a single line with comma-separated values or in multiple lines
```

just a minor type "o" should be "or"

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ClientHeaderParamTest.java
michalszynkiewicz: good catch and sorry, I forgot to `git stash` the rest of the changes when checking this commit :(

PR #169, opened 2019-03-14
Retrieve HTTP headers from parameter instead of injection.

### LINKED ISSUE #164: Injection of HttpHeaders in ClientRequestFilter does not work in Jersey
Hi!

[This injection](https://github.com/eclipse/microprofile-rest-client/blob/ffdddfe345fe88f30b4bc58ffcb55896a6edc309/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/providers/ReturnWithAllClientHeadersFilter.java#L34) does not seem to work when using Jersey and consequently the test `ClientHeaderParamTest#testHeaderNotSentWhenExceptionThrownAndRequiredIsFalse` fails. When debugging I can execute `clientRequestContext.getHeaders()` in filter method and get correct headers that the test expects.

I'm not sure if the JAX-RS specification allows this injection, so I'm posting the issue here first. Any help would be greatly appreciated!

urbim:
Fixes #164.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/providers/ReturnWithAllClientHeadersFilter.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/providers/ReturnWithAllClientHeadersFilter.java
@@ -22,8 +22,6 @@ import javax.json.Json;
 import javax.json.JsonObjectBuilder;
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientRequestFilter;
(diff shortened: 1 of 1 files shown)


andymc12:
Hi @urbim - the changes look good, but can you sign your commit? You may be able to do it by just changing the commit message.  The Eclipse Foundation requires contributors to sign the Eclipse Contributor Agreement and then to sign all commits (usually via `git commit -s`).

Thanks, Andy

urbim:
Thanks for the info, I've fixed it.

Issue #166, opened 2019-03-07
Bug with InvokeWithJsonBProviderTest myObject testGet body JSON


j4fm:
InvokeWithJsonBProviderTest.java@L94-L99

The test myObject JSON in the testGet operation is this effectively this:

{"objectName": "myObject"};"quantity": 17;"date": "2018-12-04";}

Which causes the test to fail with:

Exception with semicolon line 1 position 28

The structure of the test myObject needs to be fixed for the test to pass.

j4fm:
Can be fixed with:
`                .withBody("{" +
                             "\"objectName\": \"myObject\"," +
                             "\"quantity\": \"17\"," +
                             "\"date\": \"2018-12-04\"" +
                          "}")
`

pdudits:
After the JSON is syntactically correct, there are still few issues with mapping definition of `MyJsonBObject`.

1. JSON-B only supports instatiation over public methods and fields. Therefore not having a setter for Name causes the property not being deserialized at all.
2. For same reason, the assertion for `ignoredField` not being changed is trivially satisfied, as conforming implemention cannot set that field without public setter.
3. Deserialization of date into `java.util.Date` is bit controversial, as it is technically a timestamp format. Current version of Yasson will fail complaining that it wouldn't do that by default (See exception below). I will open an issue with them, because I believe this should work. But using `LocalDate` is much safer (and works).
4. The date in assertion does not match its expected format: `new SimpleDateFormat("yyyy.MM.dd").parse("2018-12-04")` 

This is what the exception from point 3:
````
 javax.json.bind.JsonbException: Error parsing class java.util.Date from value: 2018-12-04. 
Check your @JsonbDateFormat has all time units for class java.util.Date type, or consider using org.eclipse.yasson.YassonProperties#ZERO_TIME_PARSE_DEFAULTING.
````

Emily-Jiang:
close as it was fixed

Issue #111, opened 2018-07-31
Clarify what happens if no MIME-TYPE is defined for post/put request body


OndroMih:
If no `@Produces` and `@Consumes` are used on interface, which MIME-TYPE is used if a POST/PUT method has a body argument?

An example:

```
@Path("/root")
public interface ClientInterface {
    @PUT
    Response executePut(String body);
}
```

What MIME-TYPE will be used for body in the request? 
* Should a default of "text/plain" be used?
* Should method calls fail? With which exception?
* Should deployment fail? With which exception?

This isn't covered by the spec document. I only found a single test that uses request body and doesn't define the MIME-TYPE, `BeanParamTest`, but this test intercepts the method call with a filter and returns a fake response before the body is mapped to a request.

We need a new test to cover this scenario and assert that either we have a request encoded with an expected MIME-TYPE or that an exception is thrown.

PR #172, opened 2019-03-20
Issue 111: Define application/json as default MIME type

andymc12:
This addresses issue #111 and issue #166 


Ladicek:
I know I'm late to the party, but I would like to loudly complain that _this is in fact a breaking change_ and should have been highlighted much more prominently, ideally by bumping the version number to 2.0.

Previously, this was implementation defined, and our implementation had some sensible default (not sure what exactly it was, but things just worked). Now, I have to add `@Produces` everywhere I don't expect JSON. Sad panda.

michalszynkiewicz:
We were discussing it, I don't remember exactly how it went but I remember I thought that since we're not changing it per se but specifying something previously unspecified, maybe it doesn't need a major version bump. 
Sorry for the inconvenience.

Issue #177, opened 2019-04-10
OSGi versions of packages did not change since 1.1


pdudits:
Since release 1.1, the exported version of main package is 1.1, and versions of subpackages is 1.0.1.

This clearly does not reflect the changes in the API since May 2018.

The per-package versioning was introduced in #81, but then the version of every package needs to be carefully reviewed before every release, which is unfortunately not happening.

andymc12:
@pdudits - thanks for raising this issue.  I definitely dropped the ball on this one. 

Can you take a look at my pull request (#179) to make sure it is what you would expect?

Thanks again!

pdudits:
This is my first encounter with per-package versioning, so I'm not sure what to expect. But it is confusing to me.

In my experience, clients (and tooling) usually import packages with version ranges (so usual import for API 1.2.1 would be `[1.2,2.0)`). What problem does per-package versioning solve?

From API user's point of view, it just looks inconsistent. For example, javadoc of `ClientHeaderParam` says `@since 1.2`, whereas the exported version of package is `1.1` (after the PR is merged).  And this leads to lots of false alarms when diagnosing OSGi issues -- just like I suspected, that I'm importing wrong REST client, when in fact my real issue was #178.

andymc12:
@pdudits I agree it can be confusing, but the idea of semantic version has merit.  And if the Java EE community would follow the same rules, then I think it would've avoided the problem you are seeing in #178.  Because the CDI spec didn't make any breaking changes to consumers (only to implementors), it really should have called itself 1.3 instead of 2.0 (or at least updated the package versions in such a way). 

In MP Rest Client 1.2, we only made minor changes to the API packages (changes that break implementors, but not consumers), so OSGi bundles that were generated by tools like BND should continue to work just fine - unless you are implementing the Rest Client APIs.

Here is a brief explanation of semantic versioning used in OSGi:
https://enroute.osgi.org/FAQ/210-semantic_versioning.html

Hope this helps - and thanks again for raising this issue - I've made it a note in my personal release process - and I'll look into the overall MicroProfile project release process to see if I can make a note to update OSGi versions there too.

kenfinnigan:
@andymc12 this is done right?

andymc12:
@kenfinnigan yes - I updated it for master (1.3) and the 1.2.X-service branch.  We should probably publish a 1.2.2 release that includes these changes.

kenfinnigan:
[...]

PR #179, opened 2019-04-11
Update OSGi export package versions for 1.2 release

andymc12:
Issue #177 correctly identified that we did not update the OSGi package versions prior to the 1.2 release.  I've taken a crack at updating them now (I think we made minor updates to every package), and after this has been reviewed, I'll plan to back port this back to the 1.2-service branch, and possibly cut a new micro release.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/annotation/package-info.java
@@ -33,5 +33,5 @@
  * }
  * </pre>
  */
-@org.osgi.annotation.versioning.Version("1.0.1")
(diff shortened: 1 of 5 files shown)


andymc12:
Right - and there are rules on which digit changes:
* major = something that breaks _callers_ (i.e. deleting a public method, interface, etc.)
* minor = something that breaks _implementors_ (i.e. adding a new interface method, etc.)
* micro = pretty much everything else (i.e. new private method, updated javadoc comment, etc.).

pdudits:
As commented on the issue, I think this versioning is confusing. For example `RestClientListener` was added in version 1.2, yet version of the package is 1.1.

Issue #173, opened 2019-04-02
Java 11 module-info - javadoc build failure


andymc12:
When I build with Java 11 (`mvn clean install`), I see the following error:
```
1 error
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] MicroProfile Rest Client ........................... SUCCESS [  3.171 s]
[INFO] MicroProfile Rest Client API ....................... FAILURE [  9.107 s]
[INFO] MicroProfile Rest Client TCK ....................... SKIPPED
[INFO] MicroProfile Rest Client Specification ............. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 12.603 s
[INFO] Finished at: 2019-04-02T16:00:38-05:00
[INFO] Final Memory: 31M/69M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:jar (attach-javadocs) on project microprofile-rest-client-api: MavenReportException: Error while generating Javadoc:
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in http://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
[ERROR] 
[ERROR] Command line was: /Users/andymc/jdks/java11openjdk-j9/jdk-11.0.2+9_openj9-0.12.1/Contents/Home/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/Users/andymc/dev/andymc12/mpRestClient/microprofile-rest-client/api/target/apidocs' dir.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :microprofile-rest-client-api
```

This builds just fine with Java 8.

PR #180, opened 2019-04-11
Use source 8 for javadocs

andymc12:
This change avoid an issue documented [here](https://bugs.openjdk.java.net/browse/JDK-8212233) where the Maven Javadoc Plugin will fail when building using Java 11.

Thanks to James Carman who suggested it in gitter.


andymc12:
This resolves issue #173.

PR #187, opened 2019-04-23
SSL support, fixes #155

### LINKED ISSUE #155: Will the microprofile-rest-client support the SSL configuration?
Regarding to the[ `javax.ws.rs.client.ClientBuilder` ](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/ClientBuilder.html)which provides the setting for `keystore`, `truststore`, `HostnameVerifier` and `SSLContext`. Will the `microprofile-rest-client` support them? 

If not, could you please help to advise about the proper approach?

michalszynkiewicz:


michalszynkiewicz:
this is how an implementation in the current version of SmallRye RC might look like: https://github.com/smallrye/smallrye-rest-client/pull/77/files

The above changes were used to verify the TCK changes

michalszynkiewicz:
I had trouble making more complicated scenarios work with WireMock and created `HttpsServer` to work around it: https://github.com/eclipse/microprofile-rest-client/pull/187/files#diff-bebedbf7a3d2916c645bf81d69516352

api/src/test/java/org/eclipse/microprofile/rest/client/BuilderImpl2.java
```diff
@@ -34,10 +37,31 @@ public RestClientBuilder baseUri(URI uri) {
...
+    }
+
+    @Override
+    public RestClientBuilder trustStore(KeyStore trustStore) {
+        throw new IllegalStateException("not implemented");
+    }
+
+    @Override
+    public RestClientBuilder keyStore(KeyStore keyStore, String keystorePassword) {
+        return null;
```
kenfinnigan: Any reason this returns `null` vs `IllegalStateException`?

spec/src/main/asciidoc/ssl.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+----
+
+
+For CDI injected clients, the trust store can be specified
+with MicroProfile Config properties:
+
+- `myClient/mp-rest/trustStore` to set the trust store location.
+Can point to either a classpath resource (e.g. `classpath:/client-truststore.jks`)
+or a file (e.g. `file:/home/user/client-truststore.jks`)
+- `myClient/mp-rest/trustStorePassword` to set the
```
kenfinnigan: Do other MP specs require a password in config? Wondering if there's anything we need to do around plain text passwords in config?

spec/src/main/asciidoc/ssl.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+or a file (e.g. `file:/home/user/client-keystore.jks`)
+- `myClient/mp-rest/keyStorePassword` to set the
+password for the keystore
+- `myClient/mp-rest/keyStoreType` to set the type of the key store.
+Defaults to "JKS"
+
+
+=== SSL Context
+For the programmatically created client, it is also possible to
+configure the SSL by setting the `SSLContext` using the `RestClientBuilder#sslContext` method.
```
kenfinnigan: Remove "the" so it's just "configure SSL"

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/ssl/ClientWithKeystoreFromClasspathAndTruststore.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.rest.client.tck.interfaces.ssl;
+
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+import org.eclipse.microprofile.rest.client.tck.interfaces.JsonPClient;
+
+/**
```
kenfinnigan: Don't think we need author tags

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -165,6 +168,43 @@ default RestClientBuilder baseUri(URI uri) {
...
     RestClientBuilder executorService(ExecutorService executor);
 
+    /**
+     * Specifies the SSL context to use when creating secured transport connections to server endpoints
+     * from web targets created by the client instance that is using this SSL context.
+     *
+     * @param sslContext the ssl context
+     * @return the current builder with ssl context set
+     */
+    RestClientBuilder sslContext(SSLContext sslContext);
```
andymc12: Can you add `@since 1.3` to the new API methods?

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
```diff
@@ -165,6 +168,43 @@ default RestClientBuilder baseUri(URI uri) {
      */
     RestClientBuilder executorService(ExecutorService executor);
 
+    /**
+     * Specifies the SSL context to use when creating secured transport connections to server endpoints
+     * from web targets created by the client instance that is using this SSL context.
+     *
+     * @param sslContext the ssl context
+     * @return the current builder with ssl context set
```
andymc12: For each of the new methods, there should probably also be a `@throws` javadoc tag to indicate that a NullPointerException should be thrown if any of the params are null.

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
michalszynkiewicz: done

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
michalszynkiewicz: should I also add tests that check that?

api/src/test/java/org/eclipse/microprofile/rest/client/BuilderImpl2.java
michalszynkiewicz: a mistake, I will change it to the exception

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/ssl/ClientWithKeystoreFromClasspathAndTruststore.java
michalszynkiewicz: removed

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/AbstractSslTest.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @AfterClass
+    public static void stopHttpsServer() {
+        if (httpsServer != null) {
+            httpsServer.stop();
+        }
+    }
+
+    static String configLine(Class<?> aClass, String key, String value) {
```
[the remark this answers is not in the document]
michalszynkiewicz: done

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/SslContextTest.java
```diff
@@ -0,0 +1,78 @@
...
+        RestClientBuilder.newBuilder()
+            .baseUri(BASE_URI)
+            .sslContext(sslContext)
+            .build(JsonPClient.class)
+            .get("1");
+    }
+
+    @Test
+    public void shouldFailedMutualSslWithoutSslContext() {
+        assertThrows(ProcessingException.class, () ->
```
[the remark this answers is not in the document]
michalszynkiewicz: expected exception works well too, I will change the tests

spec/src/main/asciidoc/ssl.asciidoc
michalszynkiewicz: I don't know of any...

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/SslContextTest.java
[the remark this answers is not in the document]
michalszynkiewicz: done

api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
andymc12: It wouldn't hurt, but it's up to you - I think that will probably be the behavior for most implementations, since the JAX-RS Client SSL APIs already require it.  My main concern was that the javadocs were consistent with the JAX-RS APIs.

michalszynkiewicz: If you don't think it's needed, I'd keep it with `@throws` only.

andymc12:
@michalszynkiewicz when I ran these changes through CXF, I ran into an issue with MP Config where it would not find the configuration from the WEB-INF directory.  When I put the microprofile-config.properties file in the META-INF directory of a lib JAR inside the WAR, then it worked.  

Would you be ok making that change?  Here is an example of my change:
```
        JavaArchive jar = ShrinkWrap.create(JavaArchive.class, simpleName + ".jar")
            .addClasses(JsonPClient.class, ClientWithTruststore.class, ClientWithNonMatchingStore.class,
                ClientWithKeystoreAndTruststore.class, ClientWithKeystoreFromClasspathAndTruststore.class,
                HttpsServer.class, AbstractSslTest.class)
            .addAsManifestResource(new StringAsset(config), "microprofile-config.properties")
            .addAsManifestResource(new ClassLoaderAsset("ssl/" + clientTruststoreFromClasspath), clientTruststoreFromClasspath)
            .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
        webArchive.addAsLibrary(jar)
            .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
```

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/SslContextTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.net.ssl.SSLContext;
+import javax.ws.rs.ProcessingException;
+
+
+public class SslContextTest extends AbstractSslTest {
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        WebArchive webArchive = ShrinkWrap.create(WebArchive.class)
```
andymc12: For the tests, can you add the test class's simple name as the archive names?
```suggestion
        WebArchive webArchive = ShrinkWrap.create(WebArchive.class, SslContextTest.class.getSimpleName() + ".war")
```

michalszynkiewicz:
@andymc12 yes, I can change it. That looks a bit weird to me though. 
Does it mean that for your MP Config implementaiton, users cannot have `microprofile-config.properties` in their main project, they have to create another module with the file?

kenfinnigan:
It's been a while since I've gone into the workings of Shrinkwrap. Where does `addAsManifestResource` put the file in a WAR?

Is it in <root>/META-INF or <root>/WEB-INF/classes/META-INF ?

I found an MP Config TCK test that uses `.addAsWebInfResource(mpConfig, "classes/META-INF/microprofile-config.properties");` to add the file to a WAR.

I agree that adding the config file to the WAR should work, it's just a question of whether it's in the right place in the WAR

andymc12:
@michalszynkiewicz it definitely seems like a limitation in the Geronimo Config code, but based on a brief conversation I had with @Emily-Jiang, it appears to correct in the spec.  The MP Config code expects to be able to load the microprofile-config.properties file from the web application's class path.  That should include anything in the WEB-INF/classes directory and the JAR files in the WEB-INF/lib directory.

To @kenfinnigan's point, the MP Config TCK test works because it puts the props file in the WEB-INF/classes directory.  That should be an option for these tests too.  But, since the META-INF directory for the WAR is technically not part of the web app's class path, we should not test for that - it sounds like SmallRye and other vendors will check the WAR's META-INF directory, but that seems to be beyond the spec.

I have confirmed that if I change this `.addAsManifestResource(new StringAsset(config), "microprofile-config.properties")` to this `.addAsWebInfResource(new StringAsset(config), "classes/META-INF/microprofile-config.properties")` (matching the MP Config TCKs) then everything works in CXF.  This might be a better change than using a JavaArchive inside the WebArchive.

@michalszynkiewicz if you want, we can merge this PR as-is and then I can make the change from manifest resource to webinf resource in a separate pull.

Thanks, Andy

kenfinnigan:
+1 to using the same approach as MP Config

Issue #204, opened 2019-06-27
SSL TCK tests errors on IBM JDK


nziakova:
These 4 tests have errors on IBM JDK 1.8:
- org.eclipse.microprofile.rest.client.tck.ssl.SslContextTest.arquillianBeforeClass
- org.eclipse.microprofile.rest.client.tck.ssl.SslHostnameVerifierTest.arquillianBeforeClass
- org.eclipse.microprofile.rest.client.tck.ssl.SslMutualTest.arquillianBeforeClass
- org.eclipse.microprofile.rest.client.tck.ssl.SslTrustStoreTest.arquillianBeforeClass

```
Caused by: java.io.IOException: Invalid keystore format
	at com.ibm.crypto.provider.JavaKeyStore.engineLoad(Unknown Source)
	at java.security.KeyStore.load(KeyStore.java:1456)
	at org.eclipse.jetty.util.security.CertificateUtils.getKeyStore(CertificateUtils.java:54)
	at org.eclipse.jetty.util.ssl.SslContextFactory.loadKeyStore(SslContextFactory.java:1194)
	at org.eclipse.jetty.util.ssl.SslContextFactory.load(SslContextFactory.java:334)
	at org.eclipse.jetty.util.ssl.SslContextFactory.doStart(SslContextFactory.java:256)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:167)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:119)
	at org.eclipse.jetty.server.SslConnectionFactory.doStart(SslConnectionFactory.java:94)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:167)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:119)
	at org.eclipse.jetty.server.AbstractConnector.doStart(AbstractConnector.java:282)
	at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:81)
	at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:236)
[...]

PR #205, opened 2019-06-27
[Issue 204] Explicitly use PKCS12 store type for TCK tests

andymc12:
The SSL TCK tests use PKCS12 key/trust store types, but the tests themselves either specify JKS or rely on the default store type (for JDK 8, that is JKS - for 9+, that is PKCS12). Most JDKs can handle PKCS12 types, even when the code specifies JKS, but IBM JDK 8 cannot.

This change explicitly sets the store type in JVM system properties, the Jetty test code and MP Config properties for the test application.

This PR is still a work in progress - when finished, it should resolve issue #204.

This PR should be merged to 1.3.X-service and master.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/AbstractSslTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/AbstractSslTest.java
@@ -70,7 +70,7 @@ public abstract class AbstractSslTest extends Arquillian {
 
 
     public static KeyStore getKeyStore(File keystoreFile) throws Exception {
-        KeyStore keystore = KeyStore.getInstance("JKS");
+        KeyStore keystore = KeyStore.getInstance("pkcs12");
         try (FileInputStream input = new FileInputStream(keystoreFile)) {
             keystore.load(input, PASSWORD.toCharArray());
         }
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/HttpsServer.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ssl/HttpsServer.java
@@ -50,12 +50,14 @@ public class HttpsServer {
     public HttpsServer keyStore(String keystore, String keyStorePassword) {
         sslContextFactory.setKeyStorePath(keystore);
         sslContextFactory.setKeyStorePassword(keyStorePassword);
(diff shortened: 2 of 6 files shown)


andymc12:
@michalszynkiewicz @asoldano thanks for the review!   did you (or @nziakova) get a chance to test these changes with RestEasy and the IBM JDK 8?  These changes got us much further with testing Liberty on IBM JDK 8, but now we are seeing some issues with TLS - unable to find a common cipher.  

If the SSL tests are passing with these changes on RestEasy with IBM JDK 8, then I could go ahead and merge.

I am thinking about removing the system property code from `AbstractSslTest`.  I think those changes are unnecessary given the other changes.

asoldano:
@andymc12 unfortunately, I don't personally have access to a IBM JDK suitable for this test. Let me ask @nziakova if she can perform the verification on her environment.

nziakova:
All 4 SSL tests still fail with this update, but now with different error.

`SslTrustStoreTest`, `SslMutualTest` and `SslHostnameVerifierTest` have error in `arquillianBeforeTest`:
```
Caused by: java.security.KeyStoreException: IBMKeyManager: Problem accessing key store java.io.IOException: Keystore type is not PKCS12
 	at com.ibm.jsse2.ah$a.engineInit(ah$a.java:29)
 	at javax.net.ssl.KeyManagerFactory.init(KeyManagerFactory.java:11)
 	at org.apache.http.ssl.SSLContextBuilder.loadKeyMaterial(SSLContextBuilder.java:187)
 	at org.apache.http.ssl.SSLContextBuilder.loadKeyMaterial(SSLContextBuilder.java:208)
 	at org.jboss.resteasy.client.jaxrs.ClientHttpEngineBuilder43.build(ClientHttpEngineBuilder43.java:110)
```

And `SslContextTest.shouldSucceedMutualSslWithValidSslContext` fails with:
```
RESTEASY004655: Unable to invoke request: javax.net.ssl.SSLException: Received fatal alert: internal_error
```

Still only on IBM JDK, other JDK are OK.

andymc12:
That seems similar to what we are seeing with Liberty/CXF.  Using IBM JDK 8, even after specifying the key/trust store location, it seems like the JDK is trying to use the cacerts store that ships with the JDK instead of the specified store.  The cacerts store in the JDK is of type JKS.  This seems like a bug in the IBM JDK to me - I'm still investigating.

Issue #203, opened 2019-06-13
AsyncMethodTest#testAsyncInvocationInterceptorProvider is racey


stuartwdouglas:
org.eclipse.microprofile.rest.client.tck.providers.TLAsyncInvocationInterceptor#removeContext happens after the CompletionStage in the test is complete. This means there is no guarentee the test will actually see this method as being run, as it may not have happened yet.

reta:
Hey @stuartwdouglas, we've been seeing the same issue with CXF's MP Client implementation when running the TCK. I am wondering what MP Client implementation you are running TCK against? Thank you.

stuartwdouglas:
Quarkus, so a modified version of the Smallrye implementation.

dmatej:
Unfortunately I can see it failing again now with GlassFish 7.0.25 and 7.1.0-SNAPSHOT. It doesn't fail always, but on my laptop in some 2/3 executions.
```
AsyncMethodTest>Arquillian.run:137->testAsyncInvocationInterceptorProvider:242 expected [808] but found [0]
```
```
        <dependency>
            <groupId>org.eclipse.microprofile.rest.client</groupId>
            <artifactId>microprofile-rest-client-tck</artifactId>
            <version>3.0.1</version>
            <scope>test</scope>
        </dependency>

```

Emily-Jiang:
@jamezp @jim-krueger can you take a look at this?

dmatej:
We updated to 4.0, the test fails in around 50% of executions on my machine.

jamezp:
A challenge seems reasonable to accept here. I can see it looks fragile. Sorry this took so long to look at.

PR #207, opened 2019-07-03
[Issue 203] Fix race condition in removeContext test

andymc12:
This change recycles the thread used for the asynchronous work (after the rest client work is done) and verifies that the thread locals set by the async rest client flow have been cleared.

This should go into `1.3.X-service` and `master`.

This should resolve issue #203.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/asynctests/AsyncMethodTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/asynctests/AsyncMethodTest.java
@@ -34,7 +34,10 @@ import java.util.Map;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


andymc12:
@stuartwdouglas  @reta - could you test to verify that this fixes the issue(s) you are seeing?  I've run this several times locally (with Liberty that uses CXF) and it passes, but I never saw it fail before these changes.

kenfinnigan:
@cescoffier @asoldano @ronsigal any comments?

reta:
@andymc12 thanks a lot for the fix, I will give it a try shortly

asoldano:
@kenfinnigan , I haven't reproduced the issue, but the changes look ok.

reta:
@andymc12 I was playing with the TCK today (against CXF MP client). While the current implementation of the `AsyncMethodTest::testAsyncInvocationInterceptorProvider` was failing often, after applying the patch (this PR) I have not seen any failures anymore. Thanks again!

Issue #209, opened 2019-08-14
Exclude EL API transitive dependency


kenfinnigan:
Explicitly exclude el-api from cdi-api dependency

See eclipse/microprofile#56

PR #210, opened 2019-08-21
Issue 209: Exclude EL API dependency

andymc12:
Per eclipse/microprofile#56, excluding transitive dependency on EL API.

Resolves issue #209.


api/pom.xml
```diff
@@ -31,6 +31,13 @@
         <dependency>
             <groupId>javax.enterprise</groupId>
             <artifactId>cdi-api</artifactId>
+            <scope>provided</scope>
```
kenfinnigan: As `provided` and the exclusions are defined in the root `dependencyManagement` section, does it need to be duplicated here?

PR #223, opened 2019-10-22
Fixes #222

### LINKED ISSUE #222: InvokeWithJsonPProviderTest doesn't use @RestClient
I presume it's an omission that https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/InvokeWithJsonPProviderTest.java#L65 doesn't have `@RestClient` on the injection point for the CDI bean?

kenfinnigan:


kenfinnigan:
If this is good, I can backport to 1.3.x as well

kenfinnigan:
Thanks, will create one for 1.3 as well

Issue #231, opened 2019-10-31
Update to Jakarta EE8 APIs for MP 3.3 in Feb 2020


kenfinnigan:

andymc12:
@kenfinnigan do you think this will require a new release (i.e. MP Rest Client 1.4)? Or can we update this in a micro release (i.e. 1.3.5)?

As I understand it, the Jakarta EE8 APIs are binary compatible with the Java EE8 APIs, so other than the change to the Maven coordinates, I don't think anything else needs to change, right?

Thanks!

kenfinnigan:
technically it can be micro, but I think to be included in the umbrella they only want major/minor

So it would need to be 1.4 I believe

kenfinnigan:
Does this need to be re-opened @andymc12?

Or MP REST Client isn't doing a 1.x for MP 3.3 now?

andymc12:
@kenfinnigan I'd prefer to wait and do a 2.0 release for the June MP umbrella release.  Other than the Jakarta dependencies (2.0) the only things we have done so far in master are:
* Add `@Target` to `@RestClient` - Issue #232 
* Change in how SPI class is loaded - Issue #248
* Support injection works in `ClientHeadersFactory` - Issue #234 (PR #243 still pending)

I know you had users interested in `@Target` - would you prefer to get that released in February?  Otherwise, I don't think there is really enough content in there to justify a new release yet.  Wdyt?

kenfinnigan:
Ah, yeah I know there are some Quarkus/Kotlin users wanting the `@Target` fix.

If we can do a small minor for MP 3.3 that would be great

andymc12:
This is a breaking-change due to the change in scope for the APIs.  Previously, the APIs included Java EE 7, and used the default Maven ("compile") scope which would package the dependencies with the user-created WAR files.  After this change, the scope is now "provided" which means that the dependencies should be provided by the implementation container.

PR #233, opened 2019-11-14
Update dependencies to use Jakarta EE 8 APIs

andymc12:
This pull request resolves issue #231.

Because MP Rest Client version 1.3 previously compiled against some Java EE 7 technologies (JAX-RS 2.0 and CDI 1.2, etc.) and now are compiling against Java EE 8 / Jakarta EE 8, this requires a major version update, so the version now shows 2.0-SNAPSHOT instead of 1.4-SNAPSHOT.

This change also resolves some Java compiler warnings.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/DefaultClientHeadersFactoryImpl.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/DefaultClientHeadersFactoryImpl.java
@@ -46,7 +46,6 @@ public class DefaultClientHeadersFactoryImpl implements ClientHeadersFactory {
     private final static Logger LOG = Logger.getLogger(CLASS_NAME);
 
     private static Optional<Config> config() {
-        Config c;
         try {
             return Optional.ofNullable(ConfigProvider.getConfig());
         }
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/ext/DefaultClientHeadersFactoryImpl.java
```diff
@@ -46,7 +46,6 @@
     private final static Logger LOG = Logger.getLogger(CLASS_NAME);
 
     private static Optional<Config> config() {
-        Config c;
```
kenfinnigan: What were changes in this file for?

andymc12: After upgrading the dependencies, I started seeing these compiler warnings.  I think that they are unrelated to the upgrade, but the upgrade forced my IDE to recompile everything and made the warnings more obvious.  

I put the fixes to the compiler warnings in a separate commit to keep a clean history (one commit for the Jakarta upgrade, and one for the compiler warnings).  If you want, I can separate them further by putting them in separate PRs.

kenfinnigan: I'm a numpty, didn't notice there were two commits.

Issue #238, opened 2019-12-08
Resolver must not use load from parent first


Legion2:
The implementation of the SPI loading was copied from ConfigProviderResolver in microprofile-config, but there it was updated to resolve classloading issues. The Issues eclipse/microprofile-reactive-streams-operators#111 and eclipse/microprofile-config#390 describe the problem that also affects `org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver`.

The RestClientBuilderResolver implementation should be updated to resolve the classloading issues, eclipse/microprofile-config#450 should also be considered as related to the classloading problem.

Issue #103, opened 2018-05-29
public static void setInstance(RestClientBuilderResolver resolver)  is an implementation leak and should be dropped from the API


rmannibucau:
See eclipse/microprofile-open-api#224, this kind of method don't belong to the API and is only issues promises so should disappear from any coming release

andymc12:
Per discussion in [eclipse/microprofile-open-api#224](https://github.com/eclipse/microprofile-open-api/issues/224) the resolution of this issue should be to put the `RestClientBuilderResolver` (and any other SPI code) into a separate SPI artifact - so that vendors can implement this spec without allowing end-users to be able to call SPI methods.

rmannibucau:
Any rational to keep this method since multiple threads qhowed osgi and other envs dont need it? Others are ok to extract even if pby overkill.

andymc12:
CXF does not use the `setInstance` method, so from that implementor's standpoint, it could be removed.  It would be good to know whether other implementors use it or not.

We'd probably need to deprecate it first.

OndrejM:
It's often the only way how to set a custom resolver in an OSGi environment. I think that @Emily-Jiang has some opinion about this. In Payara, the resolver is provided via the service loader mechanism.

rmannibucau:
@OndrejM I know it is what @Emily-Jiang keeps saying but this is not true and there are OSGi solution nowdays which avoid to break the API introducing some vendor specific implementations, please have a look to https://github.com/eclipse/microprofile/issues/33 or geronimo which does it since almost 20 years without any need of API tweak.

OndrejM:
I believe, @rmannibucau. I even think that the API doesn't need to support the service loader pattern, which is also impl-specific. The SPI package exists only to make it easier to create implementations. Maybe a better solution is to provide an SPI to provide an impl-specific extension, which could provide implementations in any way

rmannibucau:
Agree, if you think 2s about here is concrete status of that particular point:

1. JavaEE always used SPI so MP "copied" that pattern but (point 2)
2. MP is CDI centric so the way to have a SPI is to use CDI beans. MP API would provide an interface and the implementation would add a bean for that type. it works for 90% of the cases but the config one which is needed before the container is started but that's really all.

Sounds like a trivial fix, no?

OndrejM:
It sounds OK to me to use CDI for providing implementations for specs that can rely on CDI. Most of them should except the config one you mentioned.

[...]

PR #239, opened 2019-12-09
238: Remove recursive classloader check in loadSpi

andymc12:
Resolves issue #238 - using the same pattern as MP Config.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/spi/package-info.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/spi/package-info.java
@@ -23,5 +23,5 @@
  * provide additional functionality for MP Rest Clients.
  *
  */
-@org.osgi.annotation.versioning.Version("1.1")
+@org.osgi.annotation.versioning.Version("1.1.1")
 package org.eclipse.microprofile.rest.client.spi;
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016-2019 Contributors to the Eclipse Foundation
  *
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/spi/package-info.java
```diff
@@ -23,5 +23,5 @@
  * provide additional functionality for MP Rest Clients.
  *
  */
-@org.osgi.annotation.versioning.Version("1.1")
```
kenfinnigan: any reason why a patch over minor?

michalszynkiewicz:
@andymc12 do you know what might have been the reason for the former approach?

api/src/main/java/org/eclipse/microprofile/rest/client/spi/package-info.java
andymc12: I was thinking about removing the `setInstance` method per issue #103 - which would force the package version to be 2.0, but then I realized that we hadn't actually come to a consensus on that.  Since this change isn't changing any method signatures, it shouldn't require a minor update. I'm not an expert in semantic versioning, but I think in this case, a micro version update is correct. 

I'd also like to separate the "spi" package into it's own artifact - that might require a version change too, but I'm not positive about that.

andymc12:
@michalszynkiewicz I'm not sure.  The previous approach was copied almost 100% from MP Config (and then changed to use RestClientResolver instead of ConfigResolver).  This change is staying in sync with the changes to MP Config (and other specs) - and it looks right to me.

The older approach was probably only useful in environments like Liberty which uses OSGi and isolates implementation classes from the application classloaders. I know that Liberty doesn't need this old functionality, so it should be safe there - it might be worth checking with other app servers to see if it works there - I suspect it is fine for Wildfly/Thorntail/Quarkus/etc. and GlassFish.  

I think TomEE uses OSGi, so it might be an issue there...  @cesarhernandezgt is this something you could try out?

api/src/main/java/org/eclipse/microprofile/rest/client/spi/package-info.java
kenfinnigan: Took a quick look at semVer and it probably does align with a patch version

struberg:
FYI: TomEE does _not_ use OSGi!

Issue #244, opened 2020-01-07
Dependencies in pom.xml for API (and parent) should use `provided` scope


andymc12:
To avoid pulling any unnecessary dependencies, we need to use `provided` scope - this should include JAX-RS, CDI, annotations, injection, etc.

PR #245, opened 2020-01-07
[244] Use provided scope for all API dependencies

andymc12:
Using provided scopes as to not repackage dependencies.  

Organizing dependency versions in parent POM.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -29,39 +29,47 @@
     <description>Typesafe Rest Client APIs for MicroProfile :: API</description>
     <dependencies>
         <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <version>${jakarta.annotation-api.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


kenfinnigan:
This can't be merged until after a possible minor release for MP 3.3 correct?

andymc12:
Yes, that's correct.  I'll hold off from merging this.  I think I may end up creating a new branch (1.4.x) - and I'll need to revert the Jakarta dependencies change in that one.  Once there is a separate branch from 1.4.x, then we should be good to merge this into master.  Thanks, Ken!

kenfinnigan:
If there's lots of things in master we wouldn't release for 1.4, maybe create 1.4 branch from 1.3 latest tag and just bring across the needed commits.

Saves reverting master

andymc12:
> If there's lots of things in master we wouldn't release for 1.4, maybe create 1.4 branch from 1.3 latest tag and just bring across the needed commits.

Good call - that's what I ended up doing.  This Pull Request should be good to go into master - and should not be backported.  The other active pull requests should go into master, but should also be backported.

andymc12:
Resolves issue #244.

Issue #249, opened 2020-02-03
Make @Context injection optional


ronsigal:
I've created "Client running in server context" (https://github.com/eclipse-ee4j/jaxrs-api/issues/842), asking for some clarification from the JAX-RS group about the relationship between a Client and its server context. I propose that @Context injection be optional until the issue is clarified.

PR #250, opened 2020-02-03
[#249] Make `@Context` injection optional

andymc12:
Per conversation in gitter, we decided that `@Context` injection of JAX-RS components should be optional for this release (we can make it required in 2.0 if we choose).  This resolves issue #249.

@ronsigal - I can't add you as a reviewer, but please take a look at the changes and approve (or add comments) as applicable.  Thanks!


ronsigal:
Thanks, @andymc12. The changes look good to me.

Issue #251, opened 2020-02-07
MP Rest Client 1.4 should depend on MP Config 1.4 (both MicroProfile 3.3)


tevans78:
The API and TCK currently have a dependency on MP Config 1.3. Now that MP Config 1.4 is final, that dependency should be updated.

https://github.com/eclipse/microprofile-rest-client/blob/a3a8069f8eb09d866970bd8c396e072809d7550f/pom.xml#L38

PR #252, opened 2020-02-07
Update MP Config dependency to 1.4

andymc12:
This is going into the 1.4 branch.  If approved, I will also make the change in the master (2.0) branch.

This resolves issue #251


andymc12:
No review, but the Config folks are telling me that this is required, and the last day to release is tomorrow, so I'm merging without review.

PR #258, opened 2020-03-09
Remove obsolete constructor target for RestClient annotation

### LINKED ISSUE #257: @RestClient - the current @Target is wrong
The current value is `@Target({ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD})`.

1. It's not possible to inject a rest client into a method or a constructor parameter (missing `ElementType.PARAMETER`).
2. It's not possible to define test mocks for rest client interfaces (missing `ElementType.TYPE`).

CDI qualifiers typically define the target as follows: `@Target({METHOD, FIELD, PARAMETER, TYPE})`. `ElementType.CONSTRUCTOR` is completely useless.

andreas-eberle:
Since, the `@RestClient` annotation makes no sense to be used on a constructor (meaning to directly annotate the constructor), I removed the `ElementType.CONSTRUCTOR`.

Fixes #257, follow up for #259.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/inject/RestClient.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/inject/RestClient.java
@@ -36,7 +36,7 @@ import java.lang.annotation.Target;
  *
  * This will cause the injection point to be satisfied by the MicroProfile Rest Client runtime.
  */
-@Target({ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE})
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE})
 @Qualifier
 @Documented
 @Retention(RetentionPolicy.RUNTIME)


gsmet:
So, to give some context, this is a bit of a problem for us as Quarkus users are using this qualifier for parameters and types.

And this is a regression compared to the previous versions.

Thanks!

phillip-kruger:
@andreas-eberle The removal of ElementType.CONSTRUCTOR is a breaking change that can only be release in June, so maybe we can split that out into a separate PR ?

phillip-kruger:
This PR can change to only include the removal of CONSTRUCTOR. I created a new PR for the addition of the other targets. So we can do a minor release without a breaking change.

See https://github.com/eclipse/microprofile-rest-client/pull/259

andymc12:
I agree with @phillip-kruger that we can add the new types for a 1.4.1 release, but we should not remove the Constructor type until the 2.0 release in June.

@andreas-eberle once Phillip's #259 goes in, you may need to rebase on his changes.  After that, I'll approve and merge this change for removing the Constructor type.  Thanks for the PR!

andreas-eberle:
@andymc12: I rebased the PR and also updated the message and title.

PR #259, opened 2020-03-09
Added ElementType.PARAMETER, ElementType.TYPE for @Target on RestClient

phillip-kruger:
Signed-off-by:Phillip Kruger <phillip.kruger@gmail.com>

PR similar to #258, except we keep CONSTRUCTOR in for now, and do the remove in a separate PR, as it's a breaking change.


andymc12:
@phillip-kruger can you also merge this to the `1.4.x-service` branch so that we could include this in a follow-up release?

phillip-kruger:
Thanks @andymc12 - see https://github.com/eclipse/microprofile-rest-client/pull/260 for the 1.4 PR

PR #266, opened 2020-04-17
[#11] Describe and test SSE Support

### LINKED ISSUE #11: Define how SSE works with the type safe clients

andymc12:
- Spec changes describing SSE Support
- TCK tests

Fixes #11

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -101,6 +101,11 @@
             <version>${jakarta.ws.rs-api.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.reactivestreams</groupId>
+            <artifactId>reactive-streams-tck</artifactId>
+            <version>1.0.3</version>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -46,23 +46,6 @@
             <scope>provided</scope>
         </dependency>
 
-        <!-- TODO: DELETE START -->
-        <!--
-        <dependency>
-            <groupId>org.apache.cxf</groupId>
-            <artifactId>cxf-rt-rs-mp-client</artifactId>
-            <version>3.3.6</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.eclipse</groupId>
-            <artifactId>yasson</artifactId>
-            <version>1.0.7</version>
-        </dependency>
-    -->
-        <!-- TODO: DELETE END -->
-
-
         <dependency>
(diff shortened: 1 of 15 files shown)


kenfinnigan:
@phillip-kruger and @ronsigal could you take a look

tck/pom.xml
```diff
@@ -39,6 +39,30 @@
...
+            <version>3.3.6</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse</groupId>
+            <artifactId>yasson</artifactId>
+            <version>1.0.7</version>
+        </dependency>
+    -->
+        <!-- TODO: DELETE END -->
```
phillip-kruger, line 63: Do you want to keep this here ?

andymc12: Good catch @phillip-kruger - at first, I was "cheating" by testing the CXF implementation in the same project as the TCK.  I've removed it.

andymc12:
As for the roadmap, here are the issues currently targeted for 2.0:
https://github.com/eclipse/microprofile-rest-client/milestone/6

I tried to sort it in order of priority - we're running close to the deadline for 2.0, so some of these issues won't make the cutoff.  Of course, others are welcome to submit pull requests for any of those issues - or suggest other issues to be included too.  

Thanks! Andy

phillip-kruger:
Seems like there is no action here. Is there anyone else that can review ?

andymc12:
There has been some review of the implementation in the CXF community - I have received some great feedback at https://github.com/apache/cxf/pull/660 - to sum up:
1) Implementations should pass the reactive streams TCK for the `Publisher<InboundSseEvent>` implementation.
2) Ideally, we would use only the Flow APIs - I tend to agree with this, but it doesn't seem like we can mandate that without increasing the minimum Java level to 11 (Flow was introduced in 9, but 9 and 10 are both out of support).
3) (More implementation-specific, but might be worth additional clarification in the spec) Should use the `ExecutorService` specified when the client instance is built or default to `ForkJoinPool.commonPool()` if none is specified.

For (1) I've added the Reactive Streams TCK into the MP Rest Client TCK and verified that the non-optional tests pass in the CXF implementation.
For (2) I think there is nothing more we can do in the spec other than wait for the MP community to increase the minimum Java level.
For (3) I'll plan to add more text to the spec indicating that SSE processing is intended to be asynchronous and should execute on the ExecutorService specified when the client instance is built.

andymc12:
I'm going to call lazy-consensus here - it has been reviewed by some folks in CXF (many thanks!), and I'd like to publish a milestone build that includes this feature in the near future.  If others review it and find problems, we can address those issues then.  Thanks!

PR #285, opened 2020-07-20
InvokeWithJson*ProviderTest: set up clients in test methods

michalszynkiewicz:


michalszynkiewicz:
The master version seems to already have clients initialized in test methods: https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/jsonb/InvokeWithJsonBProviderTest.java

PR #289, opened 2020-09-01
Clarify wording of spec to make clear @ClientHeaderParam entries are not returned

### LINKED ISSUE #288: Parameters defined via `@ClientHeaderParam` don't get applied to outbound request
Hi everybody,

when using `@RegisterClientHeaders` along with `DefaultClientHeadersFactoryImpl` I expected the factory to behave like it's described in the specification under [Specifying Additional Client Headers](https://download.eclipse.org/microprofile/microprofile-rest-client-2.0-RC2/microprofile-rest-client-2.0-RC2.html#_specifying_additional_client_headers)
> This interface has a single method and takes two read-only MultivaluedMap parameters ... The method should return a MultivaluedMap that contains the headers to merge with this second map for the "final" map of headers to be sent to the outbound processing flow.

Now, when I check the respective [code](https://github.com/eclipse/microprofile-rest-client/blob/master/api/src/main/java/org/eclipse/microprofile/rest/client/ext/DefaultClientHeadersFactoryImpl.java) the second parameter `clientOutgoingHeaders` is actually not used (except for logging) and thus not merged to the output `MultivaluedMap`:
```java
    @Override
    public MultivaluedMap<String, String> update(MultivaluedMap<String, String> incomingHeaders,
                                                 MultivaluedMap<String, String> clientOutgoingHeaders) {

        if (LOG.isLoggable(Level.FINER)) {
            LOG.entering(CLASS_NAME, "update", new Object[]{incomingHeaders, clientOutgoingHeaders});
        }
        MultivaluedMap<String, String> propagatedHeaders = new MultivaluedHashMap<>();
        Optional<String> propagateHeaderString = getHeadersProperty();
        if (propagateHeaderString.isPresent()) {
            Arrays.stream(propagateHeaderString.get().split(","))
                  .forEach( header -> {
                      if (incomingHeaders.containsKey(header)) {
                          propagatedHeaders.put(header, incomingHeaders.get(header));
                      }
                  });
        }
        if (LOG.isLoggable(Level.FINER)) {
            LOG.exiting(CLASS_NAME, "update", propagatedHeaders);
        }
        return propagatedHeaders;
    }
```
Thus, none of the header-values configured i.e. via `@ClientHeaderParam` get merged to the outbound request.

Is this behavior of `DefaultClientHeadersFactoryImpl` intended (since one can register its own Factory it's actually easy work around ... but still is surprising)?

 Regards

kenfinnigan:
Fixes #288

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ClientHeadersFactory.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/ext/ClientHeadersFactory.java
@@ -28,7 +28,9 @@ import javax.ws.rs.core.MultivaluedMap;
  * <code>HeaderParam</code> annotations on the client interface.
  * <p>
  * This method should return a MultivaluedMap of headers to be merged with the
- * outgoing headers. This will determine the final set of HTTP headers that will
+ * outgoing headers. If it's desired for {@code clientOutgoingHeaders} to be present in
+ * addition to any propagated headers, {@link #update(MultivaluedMap, MultivaluedMap) update}
+ * needs to combine the two sets to return. This will determine the final set of HTTP headers that will
  * be sent to the outbound entity provider processing chain - thus any filters,
  * MessageBodyWriters, interceptors, etc. could further refine the set of
  * headers actually sent on the client request.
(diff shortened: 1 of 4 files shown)


kenfinnigan:
I've modified the PR to just focus on clarifying the wording and adding a test to verify that the headers passed in are not returned

andymc12 (approved):
LGTM - thanks for the clarification

kenfinnigan:
@andymc12 do we do an RC or wait?

andymc12:
@kenfinnigan I'd say wait.  Once the WG gets ironed out, then we can release the final GA version of 2.0 that will include these changes.  Since it only clarifies things in the spec/javadoc, it shouldn't have any impact on users/implementors.  Thanks again for the changes!

PR #322, opened 2021-07-28
ReactiveStreamsPublisherTckTest - setup cleanupLatch only when starting server

aubi:
The test was waiting for cleanupLatch, although it wasn't always used,
which caused timeouts:
[ReactiveStreamsPublisherTckTest] [ERROR] Server did not close long after test completed

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/ReactiveStreamsPublisherTckTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/ReactiveStreamsPublisherTckTest.java
@@ -15,9 +15,6 @@
  */
 package org.eclipse.microprofile.rest.client.tck.sse;
 
-import static org.eclipse.microprofile.rest.client.tck.sse.AbstractSseTest.launchServer;
-import static org.eclipse.microprofile.rest.client.tck.sse.AbstractSseTest.PORT;
-
 import java.lang.reflect.Method;
 import java.net.URI;
 import java.security.AccessController;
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/ReactiveStreamsPublisherTckTest.java
```diff
@@ -26,10 +23,10 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-
 import javax.ws.rs.sse.InboundSseEvent;
-
 import org.eclipse.microprofile.rest.client.RestClientBuilder;
+import static org.eclipse.microprofile.rest.client.tck.sse.AbstractSseTest.PORT;
+import static org.eclipse.microprofile.rest.client.tck.sse.AbstractSseTest.launchServer;
```
andymc12, line 29: Would you undo the import order changes?  My guess is that your IDE did that for you automatically - but there may be a switch to disable that save-action.

andymc12 (changes requested):
Sorry for the delayed review.  I think this looks good, but would you please undo the import changes and also update the copyright?  It should change from:
```
Copyright (c) 2020 Contributors to the Eclipse Foundation
```

to:
```
Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
```

After that I should be able to merge it.  Thanks!

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/ReactiveStreamsPublisherTckTest.java
aubi: Exactly as you said. Reverted.

andymc12 (approved):
Thanks so much!

PR #329, opened 2021-09-01
Update config dependency

Emily-Jiang:


andymc12 (approved):
Thanks for updating the Config dependency, @Emily-Jiang!

PR #331, opened 2021-09-21
Fix license

Emily-Jiang:


tck/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,12 @@
+=========================================================================
+==  NOTICE file corresponding to section 4(d) of the Apache License,   ==
+==  Version 2.0, in this case for Microprofile Config                  ==
+=========================================================================
+
+SPDXVersion: SPDX-2.1
+PackageName: Eclipse Microprofile
```
andymc12, line 7:
```suggestion
PackageName: Eclipse MicroProfile
```

Not sure if it matters here, but I usually see the "P" capitalized.

Emily-Jiang:
@andymc12 this change is to add the correct license to the TCK jars. Because the parent pom relies on the empty file tck under META-INF/resources under TCK folder, we need to create that. For Release candidates, we need to package the apache license. This PR is purely required by microprofile-parent pom.xml so that the right license is pulled into the tck jar.

PR #333, opened 2021-09-24
[330] Skip Conv-scope test in non-servlet env

### LINKED ISSUE #330: TCK Challenge: HasConversationScopeTest
Since `@ConversationScoped` is optional for an MP platform implementation, I suggest moving the following TCK test `org.eclipse.microprofile.rest.client.tck.cditests.HasConversationScopeTest`  also to an optional suite, so it is not required for an implementation to pass such test if it doesn't support `@ConversationScoped`

andymc12:
Fixes #330 - this will be backported to the `jakarta91-branch` branch (and `2.0.X-service` branch if requested).

This should skip the `HasConversationScopeTest` tests if no servlet APIs are available.

@radcortez can you verify that your non-servlet implementation passes(skips) these tests with this fix?  Thanks!

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/HasConversationScopeTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/HasConversationScopeTest.java
@@ -18,6 +18,8 @@
 
 package org.eclipse.microprofile.rest.client.tck.cditests;
 
+import static org.eclipse.microprofile.rest.client.tck.utils.ClassUtils.canLoad;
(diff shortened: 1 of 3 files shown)


radcortez:
> @radcortez can you verify that your non-servlet implementation passes(skips) these tests with this fix? Thanks!

This is a good idea and indeed works. On the other hand, I think it is fragile in the sense that there is a high chance that the servlet jar is included transitively by other dependencies, making it think that there is an implementation where there is none. Looking for the implementations is not really scalable.

I guess the only real safe solution is to separate into a different test suite, so we can have a full suite and a microprofile only suite. Do you have any other ideas?

PR #339, opened 2021-11-18
prepare for the final release of Rest Client 3.0

Emily-Jiang:


api/src/test/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,12 @@
+=========================================================================
+==  NOTICE file corresponding to section 4(d) of the Apache License,   ==
+==  Version 2.0, in this case for Microprofile Config                  ==
```
andymc12, line 3: Rest Client instead of Config?

```suggestion
==  Version 2.0, in this case for Microprofile Rest Client             ==
```

andymc12 (changes requested):
Thanks for the changes @Emily-Jiang - my only concern is the use of "Config" in the NOTICE file.  Is this intentional, or should it be "Rest Client"?

Emily-Jiang:
Thanks @andymc12 ! It was a typo. Fixed it now.

api/src/test/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,12 @@
+=========================================================================
+==  NOTICE file corresponding to section 4(d) of the Apache License,   ==
+==  Version 2.0, in this case for Microprofile Rest Client             ==
```
Emily-Jiang, line 3:
```suggestion
==  Version 2.0, in this case for MicroProfile Rest Client             ==
```

andymc12 (approved):
LGTM - thanks @Emily-Jiang!

PR #354, opened 2023-03-14
[353] Use a privileged action for iterating over the service loader w…

### LINKED ISSUE #353: RestClientBuilderResolver does not work when the security manager is enabled
The `RestClientBuilderResolver` does some privileged actions for various things like looking up the class loader. However, the `loadSpi()` uses a `ServiceLoader` and does not use a privileged action requiring ` java.io.FilePermission` to be added to read the SPI JAR.

jamezp:
…hen locating the RestClientBuilderResolver SPI.

resolves #353 

It would be nice if we could get this in 3.0 as well. If that seems acceptable, I'm happy to open up another PR.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
@@ -51,7 +51,7 @@ public abstract class RestClientBuilderResolver {
      * Implementations are expected to override the {@link #newBuilder()} method to create custom RestClientBuilder
      * implementations.
      * <p>
-     * 
+     *
      * @return new RestClientBuilder instance
      */
     public abstract RestClientBuilder newBuilder();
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
```diff
@@ -95,19 +95,21 @@ private static RestClientBuilderResolver loadSpi(ClassLoader cl) {
...
+                }
+                instance = spi;
             }
-            instance = spi;
-        }
+            return instance;
+        };
 
-        return instance;
+        return System.getSecurityManager() == null ? action.run() : AccessController.doPrivileged(action);
```
rdebusscher, line 112: It is not wise to introduce a new usage of SecurityManager which is Deprecated for removal in JDK 18. This limits the usability in an LTS JDK21+.

rdebusscher (changes requested):

api/src/main/java/org/eclipse/microprofile/rest/client/spi/RestClientBuilderResolver.java
jamezp: There are users that require the security manager and this already has references to the security manager. It was just missed in this one spot.

Emily-Jiang: We can merge this one and then clean up all SecurityManager usage when it is removed from JDK in due course.

Emily-Jiang (approved):
LGTM

Issue #366, opened 2024-01-02
No TCK coverage for exception mappers when subresource interface methods are called


cghislai:
I noticed that the resteasy implementation did not call a registered ResponseExceptionMapper when subresource interface methods are called; while it works fine for normal resources.

I suggest adding a test for this use-case; and maybe other providers could be tested on subresources as well.

See https://github.com/resteasy/resteasy-microprofile/issues/239#issue-2058665792 for the resteasy issue.

PR #367, opened 2024-01-02
Add a test for proper handling of ResponseExceptionMapper providers on subresource interface method calls

cghislai:
Fixes issue #366


jamezp:
@cghislai Thank you! I will validate this against RESTEasy MicroProfile since the fix is now merged.

jamezp (approved):
I can confirm this passes on resteasy-microprofile as well.

Emily-Jiang (approved):
LGTM

PR #368, opened 2024-01-21
[360] Upgrade the parent to work with newer versions of the Java comp…

### LINKED ISSUE #360: Align with Jakarta EE 10
The latest version of MP Rest Client 3.0 and 3.0.1 are based on the microprofile-parent pom version 2.6, which depends on Jakarta EE 9 API dependencies.

Therefore the OSGi manifest specifies Jakarta EE 9 versions of required Jakarta EE API packages and excludes versions in Jakarta EE 10. For example, the package `jakarta.enterprise.util` from CDI is required with version="[3,4)", and thus excludes CDI 4.0, which is in Jakarta EE 10. 

If MP Rest Client 3.0 requires Jakarta EE 10, then the API artifact should be based on a new version of microprofile-parent pom, which requires Jakarta EE 10 dependencies.

If MP Rest Client 3.0 should work with both Jakarta EE 9 and 10, then the OSGi manifest needs to be modified to include also newer versions of dependencies in Jakarta EE 10 (e.g. `jakarta.enterprise.util` with version="[3,5)")

jamezp:
…iler. Use the Jakarta EE 10 BOM to align the components with Jakarta EE 10.

Part of the parent upgrade is upgrading the formatter and impsort plugins. These upgrades are the reasons for the format changes.

resolves #360

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -37,6 +37,10 @@
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>jakarta.inject</groupId>
+            <artifactId>jakarta.inject-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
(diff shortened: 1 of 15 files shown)


Emily-Jiang (approved):
LGTM! Thank you @jamezp !

PR #371, opened 2024-05-13
[348] Clarify that when a client injected as a CDI bean falls out of …

### LINKED ISSUE #348: Clarification about closing the rest client
In the [Lifecycle of Rest Client](https://download.eclipse.org/microprofile/microprofile-rest-client-3.0/microprofile-rest-client-spec-3.0.html#lifecycle) section of the spec, it's mentioned that the client should be closed after use. All the examples in the spec are for  clients that are being created via builder, not the injected one. So as a reader my assumption is that for automatically injected clients (i.e. via `@RestClient`) the close method would be called automatically when the client is not in use anymore (e.g. when the CDI bean it's injected into is not used anymore which is the same time that the `@PreDestroy` callback is called).

Unfortunately seems that's not the case. 

It seems that the injected client is not being closed automatically by the container and shortly after it's being used, I get the following in the WildFly log:

`WARN  [org.jboss.resteasy.client.jaxrs.i18n] (Finalizer) RESTEASY004687: Closing a class org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient43Engine instance for you. Please close clients yourself.`

It's not clear to me if it's a bug in the resteasy implementation or it's an expected behavior based on their interpretation of the spec. So I decided to ask here first **what is the expected behavior?**.

Do we have to implement the `@PreDestroy` callback on all such beans to close the clients injected into them?

My test setup is very simple:

**The REST client interface:**
```java
@RegisterRestClient(baseUri = "https://reqbin.com")
public interface ReqBinAPI extends AutoCloseable {
    @GET
    @Produces(MediaType.APPLICATION_JSON)
    @Path("/echo/get/json")
    String get();
}
```

**The JAX-RS resource that is calling it:**
```java
@Path("/test")
public class TestResource {

    @Inject
    @RestClient
    ReqBinAPI reqBinAPI;

    @GET
    public Response method1() {
        String result = "default-value";
        try {
            System.out.println("method1 called");
            result = reqBinAPI.get();
        } catch (Exception e) {
            System.out.println(e.getMessage());
        }
        return Response.ok(result).build();
    }
}
```

**Environment:**
- WildFly 26.0.0 that is shipped with:
  - MicroProfile 2.0
  - Resteasy Client MicroProfile 4.7.4.
- OpenJDK 1.8
- Ubuntu 22.04

jamezp:
…scope, the client is closed. Add TCK tests to assert this behavior.

resolves #348

### DIFF
--- a/spec/src/main/asciidoc/lifecycle.asciidoc
+++ b/spec/src/main/asciidoc/lifecycle.asciidoc
@@ -23,6 +23,8 @@ When closed, a client instance is expected to throw an `IllegalStateException` w
 
 When a client instance is closed, the implementation is expected to clean up any underlying resources.
 
+=== Builder API
+
 A client instance can be closed by casting the instance to a `Closeable` or `AutoCloseable` and invoking the the `close()` method (or auto-invoked if using in a try-with-resources block).
 For example:
 
@@ -67,3 +69,7 @@ try (MyServiceClient c = client) {
 
 String response2 = client.greet(); // throws IllegalStateException
 ----
+
+=== Injected Client
+
+Implementations must close a client once the CDI bean falls out of scope.
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/ClientClosedTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/ClientClosedTest.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2024 Contributors to the Eclipse Foundation
+ *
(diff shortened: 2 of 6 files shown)


WhiteCat22 (approved):

WhiteCat22 (approved):

jim-krueger (approved):

WhiteCat22 (changes requested):
copyright dates

jamezp:
PR updated per the review we had on the call.

Emily-Jiang (approved):

Issue #334, opened 2021-09-27
Have string overloaded methods for uri/url methods


maxandersen:
i.e. `.baseUri("https://restcountries.eu/rest")` should be sufficient not require to always type `.baseUri(URI.create("https://restcountries.eu/rest"))`

Emily-Jiang:
This involves API, Spec and TCKs updates. @whitecat22 will work on this.

WhiteCat22:
I will take this issue.

WhiteCat22:
https://github.com/eclipse/microprofile-rest-client/pull/372

PR #372, opened 2024-05-24
[334] - add String overloaded RestClientBuilder.baseUri() method

WhiteCat22:
Added a `String` overloaded `RestClientBuilder.baseUri()` method to the API.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/rest/client/RestClientBuilder.java
@@ -101,6 +101,26 @@ public interface RestClientBuilder extends Configurable<RestClientBuilder> {
     }
 
     /**
+     * Specifies the base URI to be used when making requests. Assuming that the interface has a
+     * <code>@Path("/api")</code> at the interface level and a <code>uri</code> is given with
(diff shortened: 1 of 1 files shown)


jamezp (approved):

jim-krueger (approved):

Emily-Jiang (commented):
Can you also add a TCK for this API change? Can you also check to see whether you can make a spec change as well?

jim-krueger:
@Emily-Jiang Adam is already considering TCK for this and talking with James and me about it in slack.   

It will probably be best going forward to not consider that a PR is ready to merge until we directly request that you merge it.
Thanks

WhiteCat22:
I've finally found time to talk to both Jim and James and the consensus is that a TCK test isn't really appropriate, since the implementation is in the API. A TCK test would just be testing the JDK's implementation on `URI.create()` rather than the MpRestClient implementation.

Additionally, I don't think a spec change is appropriate either since we don't go into specific details for other API methods. We aren't adding new functionality that needs to be documented. We're just adding additional options for existing functionality to the API which is clearly detailed by the javadoc.

jamezp:
Agreed, I don't see a need for a TCK test for a single method.

jim-krueger:
@Emily-Jiang This is ready to be merged.

WhiteCat22:
resolves #334"

PR #373, opened 2024-06-05
Use the correct paths to trigger CI.

jamezp:
Note on the second commit. For some reason the `formatter-maven-plugin` was not being invoked by default. ~~It seems after #374 being merged, it started being invoked.~~

I did checkout a tag and apparently I'm wrong, this has been enabled the whole time. I don't know why this didn't get invoked for the #371 PR. Possibly it was out of date with main, I'm not sure.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/ClientClosedTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/ClientClosedTest.java
@@ -18,12 +18,6 @@
 
 package org.eclipse.microprofile.rest.client.tck.cditests;
 
-import jakarta.enterprise.context.spi.CreationalContext;
-import jakarta.enterprise.inject.spi.Bean;
-import jakarta.enterprise.inject.spi.BeanManager;
-import jakarta.inject.Inject;
-import jakarta.ws.rs.ApplicationPath;
-import jakarta.ws.rs.core.Application;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 import org.eclipse.microprofile.rest.client.tck.interfaces.cdi.scoped.AutoCloseableClient;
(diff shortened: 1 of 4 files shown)


tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/ClientClosedTest.java
```diff
@@ -18,12 +18,6 @@
 
 package org.eclipse.microprofile.rest.client.tck.cditests;
 
-import jakarta.enterprise.context.spi.CreationalContext;
-import jakarta.enterprise.inject.spi.Bean;
```
jim-krueger, line 22: Nothing is really changed in this file correct?

tck/src/main/java/org/eclipse/microprofile/rest/client/tck/interfaces/cdi/scoped/StringClient.java
```diff
@@ -16,11 +16,12 @@
```
jim-krueger, line 16: Same comment.

jim-krueger (commented):
See minor comments.

jamezp:
@jim-krueger For some unknown reason in #371, these files did not get formatted by the `formatter-maven-plugin`. The changes are needed so we don't see uncommitted changes every time you build the project. This will not happen in the future now that CI is actually enabled via this PR as well.

jim-krueger:
Yes, I saw that too and wondered why.   I'm all for not seeing them again.   Thanks

jim-krueger (approved):

jamezp:
@Emily-Jiang Could we get this merged? This will fix CI to actually start jobs and avoid further compile issues.

Emily-Jiang (approved):

PR #375, opened 2024-06-10
Bump the version to 4.0.

jamezp:
This was previously missed when we were required to move from 3.1 to 4.0


Emily-Jiang:
@jamezp you don't need to do this PR as once it is released. The code base will be reset and we can directly go to 4.1-SNAPSHOT.

jamezp:
@Emily-Jiang Ah, so when we do the release it will end up with 4.0?

PR #382, opened 2024-06-15
Clarify specification text for CDI support

### LINKED ISSUE #381: Clarify specification text for CDI support
Currently, the specification states:

> Rest Client interfaces may be injected as CDI beans. The runtime must create a CDI bean for each interface annotated with RegisterRestClient. The bean created will include a qualifier @RestClient to differentiate the use as an API call against any other beans registered of the same type. Based on the rules of how CDI resolves bean, you are only required to use the qualifier if you have multiple beans of the same type.

However, the last part is not how CDI works, the `@RestClient` annotation is not optional. We should either curate the sentence or remove it. Below is an explanation of why it's not true with links to the CDI specification. I'll send a PR shortly.

Every bean has a set of types and a set of qualifiers and because all rest client beans always contain `@RestClient`, they will not contain the [built-in `@Default` qualifier](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#builtin_qualifiers).

Now, consider following injection point:

```
@Inject
Foo foo;
```

To satisfy this injection point, you need a bean of type `Foo` but also with`@Default` qualifier ([2.2.8. The default qualifier at injection points](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#injection_point_default_qualifier)). As shown above, this is a condition no rest client bean satisfies.

FTR, this originally came up in a [Quarkus issue](https://github.com/quarkusio/quarkus/issues/41188#issuecomment-2167914024) where there was a discussion about possibly omitting the qualifier.

manovotn:
Fixes #381 

See linked issue for detailed explanation.
In short, The `@RestClient` qualifier is not optional and its presence is there to make sure rest client beans don't clash (and cause ambig. dependency resolution) with any other CDI bean with the same type.

This isn't changing current behavior, it is merely a removal of an incorrect explanation of CDI resolution rules.

### DIFF
--- a/spec/src/main/asciidoc/cdi.asciidoc
+++ b/spec/src/main/asciidoc/cdi.asciidoc
@@ -17,7 +17,7 @@
 [[restcdi]]
 == MicroProfile Rest Client CDI Support
 
-Rest Client interfaces may be injected as CDI beans.  The runtime must create a CDI bean for each interface annotated with `RegisterRestClient`.  The bean created will include a qualifier `@RestClient` to differentiate the use as an API call against any other beans registered of the same type.  Based on the rules of how CDI resolves bean, you are only required to use the qualifier if you have multiple beans of the same type.  Any injection point or programmatic look up that uses the qualifier `RestClient` is expected to be resolved by the MicroProfile Rest Client runtime.  Below is an example of said interface, with its matching injection point:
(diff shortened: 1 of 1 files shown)


Emily-Jiang (approved):
LGTM! Thanks @manovotn

PR #384, opened 2024-06-18
consome the parent pom 3.3

Emily-Jiang:


jamezp (approved):
Approving with an acknowledgement that we will do a follow up PR to continue testing with Java SE 11. The `formatter-maven-plugin` now requires Java SE 17 as a minimum which is requires compiling with SE 17.

PR #387, opened 2024-06-26
Follow up TCK fixes for two types missing from deployments.

jamezp:
These were not previously caught because the TCK in the RESTEasy MicroProfile project was run in embedded Jetty. However, these need to be fixed to run other containers which truly isolate deployments.


jim-krueger (approved):
Looks good!!

WhiteCat22 (approved):

WhiteCat22:
Thanks James!

PR #406, opened 2026-01-07
Add automatic module name to JAR

### LINKED ISSUE #397: Please add `Automatic-Module-Name`
Java 9 introduced modules. To depend on non-module code from modules, there are two options:
- JAR includes `Automatic-Module-Name` in its META-INF/MANIFEST.MF
- Name autogenerated from the filename

The latter is obviously not great because it can change at any time. Please add the `Automatic-Module-Name` entry to the JARs published from this project.

More information: http://branchandbound.net/blog/java/2017/12/automatic-module-name/

dmlloyd:
Fixes #397

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -64,4 +64,19 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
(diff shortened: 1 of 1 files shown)


jamezp:
Is there a reason not to use a `module-info.java`? I guess we end up having to wait for MicroProfile Config to be released first.

jamezp:
On the MicroProfile Technical call today, it was decided that we will first release MicroProfile Config, then create a `module-info.java` instead of using an automatic module name.

jamezp (approved):
LGTM.

dmlloyd:
Any further word on this one?

jamezp:
I'm not a committer so there isn't much I can do. There is #409 now too which looks okay to me.

